PR #118, opened 2026-05-26
feat: replace pg_cron auto-cancellation with lazy evaluation (KIM-366)

Oiranca:
## Summary

- Removes pg_cron scheduled job that cancelled expired pending reservations
- Replaces with lazy evaluation: expired pending reservations (created_at + 60 min < now) are filtered out at query time in the service layer
- Cron endpoint (/api/cron/cancel-pending) is deprecated and now returns 410 Gone
- Migration removes the pg_cron job from the database

## Changes

- `app/api/cron/cancel-pending/route.ts` — returns 410 Gone (deprecated stub)
- `lib/server/reservations-service.ts` — lazy eval filter at 4 query sites; expiry anchored to `created_at`
- `supabase/migrations/20260526000001_remove_cron_lazy_eval_cancellation.sql` — removes pg_cron job
- `__tests__/` — updated tests to match new behavior

## Test plan

- [x] Build passes (pnpm run build)
- [x] Type check passes
- [x] Lint passes
- [x] All 4 lazy eval tests pass (expired excluded, valid included, boundary correct, active always included)
- [x] cancel-pending tests pass (410 Gone response)
- [x] No regression in existing reservation service tests (72/72 pass)

Closes KIM-366

### DIFF
--- a/lib/server/reservations-service.ts
+++ b/lib/server/reservations-service.ts
@@ -269,9 +269,9 @@ async function listActiveReservationsForConflict(input: {
   const nowUtc = await getDatabaseNow(admin)
   return (data ?? []).filter((row) => {
     if (row.status === 'pending' && row.activated_at === null) {
-      const reservationStart = zonedDateTimeToUtc(row.date, normalizeTime(row.start_time))
-      const expiresAt = new Date(reservationStart.getTime() + GRACE_PERIOD_MINUTES * 60 * 1000)
-      return nowUtc < expiresAt // Keep only non-expired pending reservations
+      const createdAt = new Date(row.created_at)
+      const expiresAt = new Date(createdAt.getTime() + GRACE_PERIOD_MINUTES * 60 * 1000)
+      return nowUtc.getTime() <= expiresAt.getTime() // Keep only non-expired pending reservations
     }
     return true // Keep active reservations
   }) as ReservationRow[]
@@ -313,9 +313,9 @@ async function listOverlappingReservationIds(input: {
     const nowUtc = await getDatabaseNow(admin)
     const filteredRows = rows.filter((row) => {
       if (row.status === 'pending' && row.activated_at === null) {
-        const reservationStart = zonedDateTimeToUtc(row.date, normalizeTime(row.start_time))
-        const expiresAt = new Date(reservationStart.getTime() + GRACE_PERIOD_MINUTES * 60 * 1000)
-        return nowUtc < expiresAt
+        const createdAt = new Date(row.created_at)
+        const expiresAt = new Date(createdAt.getTime() + GRACE_PERIOD_MINUTES * 60 * 1000)
+        return nowUtc.getTime() <= expiresAt.getTime()
       }
       return true
     })
@@ -514,9 +514,9 @@ export async function listVisibleReservations(input: {
     .filter((row) => {
       // Lazy evaluation: treat expired pending reservations as cancelled
       if (row.status === 'pending' && row.activated_at === null) {
-        const reservationStart = zonedDateTimeToUtc(row.date, normalizeTime(row.start_time))
-        const expiresAt = new Date(reservationStart.getTime() + GRACE_PERIOD_MINUTES * 60 * 1000)
-        if (nowUtc >= expiresAt) {
+        const createdAt = new Date(row.created_at)
+        const expiresAt = new Date(createdAt.getTime() + GRACE_PERIOD_MINUTES * 60 * 1000)
+        if (nowUtc.getTime() > expiresAt.getTime()) {
           return false // Exclude expired pending reservations
         }
       }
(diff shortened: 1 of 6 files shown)


lib/server/reservations-service.ts
```diff
@@ -487,13 +508,27 @@ export async function listVisibleReservations(input: {
...
 
   const isAdmin = input.session.role === 'admin'
-  return (data ?? []).map((row) => {
-    const reservation = mapEnrichedReservation(row)
-    if (!isAdmin) {
-      reservation.memberNumber = undefined
-    }
-    return reservation
-  })
+  const nowUtc = await getDatabaseNow(supabase)
```
Oiranca, line 511: `getDatabaseNow(supabase)` runs through the request-scoped user client here. `supabase/migrations/20260417000004_fix_function_search_paths.sql` revokes `get_database_time()` from both `anon` and `authenticated`, so normal signed-in reservation listing will 500 in production. The same issue appears in `checkUserSlotOverlap()`. Please use an admin client for the time RPC, or avoid the RPC from user-scoped flows.

lib/server/reservations-service.ts
```diff
@@ -487,13 +508,27 @@ export async function listVisibleReservations(input: {
...
+
+  return (data ?? [])
+    .filter((row) => {
+      // Lazy evaluation: treat expired pending reservations as cancelled
+      if (row.status === 'pending' && row.activated_at === null) {
+        const createdAt = new Date(row.created_at)
```
Oiranca, line 517: This changes the expiry anchor from the reservation start time to `created_at`. The removed SQL cancelled pending rows after `date + start_time + grace_minutes`; this now expires any future reservation 60 minutes after creation. A member booking tomorrow would have the reservation hidden long before the slot starts. Please confirm the intended business rule and either preserve start-time-based expiry or update every affected flow/test to cover the new semantics explicitly.

lib/server/reservations-service.ts
```diff
@@ -487,13 +508,27 @@ export async function listVisibleReservations(input: {
...
-    if (!isAdmin) {
-      reservation.memberNumber = undefined
-    }
-    return reservation
-  })
+  const nowUtc = await getDatabaseNow(supabase)
+
+  return (data ?? [])
+    .filter((row) => {
+      // Lazy evaluation: treat expired pending reservations as cancelled
```
Oiranca, line 515: This only hides expired pending reservations in selected read/conflict paths. The row remains `status = 'pending'`, while mutation/check-in paths such as `updateReservationForSession()` and `activateReservationByTable()` still load raw pending rows without applying the same expiry guard. A user holding the id from the create response can still mutate or activate an expired hold. Please enforce the same expiry rule on those mutation/access paths, or persist `cancelled` before those flows run.

lib/server/reservations-service.ts
```diff
@@ -298,8 +307,20 @@ async function listOverlappingReservationIds(input: {
       serviceError('Internal server error', 500)
     }
 
-    const rows = data ?? []
-    reservationIds.push(...rows.map((row) => row.id))
+    const rows = (data ?? []) as ReservationRow[]
+
+    // Lazy evaluation: filter out expired pending reservations
+    const nowUtc = await getDatabaseNow(admin)
```
Oiranca, line 313: `nowUtc` is fetched inside the pagination loop, so different pages can be filtered against different timestamps if a row crosses the grace boundary mid-scan. This also adds one RPC per page and fetches full `RESERVATION_COLUMNS` when only `id`, `status`, `activated_at`, and the expiry anchor are needed. Please fetch `nowUtc` once before the loop and reduce the projection.

Oiranca (commented):
Two review cycles completed across security, correctness/performance, and QA.

Blocking issues found. I cannot submit `REQUEST_CHANGES` because GitHub rejects request-changes reviews on my own PR, so leaving this as review comments instead.

Oiranca:
Additional blocker from the second QA pass:

`lib/server/tables-service.ts` availability still appears to forward raw `active`/`pending` reservations into availability building. Since this PR stops physically cancelling expired pending rows, those stale rows can keep blocking table/room slots in the UI unless availability applies the same lazy-expiry rule. Please add coverage for expired pending reservations in table/room availability as part of this change.

Oiranca:
Thanks for catching this. Added lazy-expiry filter to `lib/server/tables-service.ts` in commit `ee8553b`.

**Changes:**
- Added `PENDING_EXPIRY_MINUTES = 60` constant to mirror grace period in reservations-service
- Modified `getTableAvailability()` to filter out pending reservations where:
  - `activated_at IS NULL`
  - `created_at + 60min <= NOW_UTC`
- Added `activated_at` to the select columns
- Parallelized `getDatabaseNow()` call with existing queries (no latency impact)

**Test coverage added (commit `76860ee`):**
- Test A: Expired pending (70min old, not activated) does NOT block availability ✓
- Test B: Active reservations still block (regression) ✓
- Test C: Fresh pending (5min old) still blocks within grace period ✓
- Test D: Activated pending is never expired ✓
- Test E: Boundary at 60min (inclusive `<=`) ✓
- Test F: Mixed scenarios with multiple states ✓

**Validation:** 23/23 tables-service tests pass + 72+ reservation tests pass. Build, typecheck, and lint all green.

This ensures stale pending reservations no longer phantom-block table/room slots in the UI.

PR #122, opened 2026-06-17
Add admin event scheduling and reservation blocking

Oiranca:
## Summary

Implements KIM-383 by adding administrator-managed events that can block one or more rooms across one or more scheduled time ranges.

## What changed

- Added database tables for events, event-room assignments, and event schedules.
- Added admin-only event CRUD endpoints under `/api/events`.
- Added an Events tab to the admin dashboard with create, edit, list, and delete flows.
- Supports selecting multiple rooms per event.
- Supports multiple date/time schedules per event.
- Integrates event occupancy into table and room availability checks.
- Blocks new or edited reservations when their room overlaps an event schedule.
- Added domain types, API endpoint helpers, TanStack Query hooks, and i18n copy.
- Added test coverage for event-blocked availability and reservation conflicts.

## Impact

Admins can now reserve room capacity for events without using normal member reservations. Members see affected slots as unavailable, and backend validation prevents bypassing the UI by posting overlapping reservations directly.

## Validation

Local push hook and manual checks passed:

- `pnpm typecheck`
- `pnpm lint`
- `pnpm test` (190/190)
- `pnpm build`

Note: the existing local security warning about `NEXT_PUBLIC_APP_URL` being unset appears during middleware tests, but the suite passes.

PR #123, opened 2026-06-17
feat(events): multi-day schedules for admin event management (KIM-383)

Oiranca:
## Summary

Implements KIM-383 by **extending the existing** events feature on `develop` (`events` + `event_room_blocks`) to support multiple date/time schedules per event (multi-day). Built fresh from `develop`.

> Replaces PR #122, which was branched from a stale `main` (425 commits behind) and reimplemented the feature from scratch with conflicting table names (`event_rooms`/`event_schedules`) and `public.is_admin()`. That branch is abandoned.

## What changed

- New migration `20260617000001_kim383_multi_day_events.sql` (timestamp after develop's latest `20260602000008`): adds `create_event_with_blocks` / `update_event_with_blocks` SECURITY DEFINER RPCs (service_role only), looping over a `jsonb` blocks array and cancelling overlapping reservations per block.
- Service layer: `createEvent`/`updateEvent` dispatch to the new RPCs when a `schedules` array is provided; legacy single-block path preserved; `deleteEvent` is multi-date aware.
- Types: `AdminEventSchedule` + `schedules: AdminEventSchedule[]` on `AdminEvent`.
- Hooks: `useAdminCreateEvent`/`useAdminUpdateEvent` accept optional `schedules`.
- Admin UI: dynamic schedule list (add/remove per-block rows); list view shows date-range + multi-day badges.
- i18n: new keys with full en/es parity (correct Spanish accents, incl. "Añadir horario").

## Validation

- `pnpm typecheck` — pass
- `pnpm test` — 566 tests (552 existing + 14 new multi-day tests)
- `pnpm build` — clean

### DIFF
--- a/components/admin/events-section.tsx
+++ b/components/admin/events-section.tsx
@@ -2,7 +2,7 @@
 
 import { useState } from 'react'
 import { useTranslations } from 'next-intl'
-import { CalendarRange, Plus, Pencil, Trash2 } from 'lucide-react'
+import { CalendarRange, Plus, Pencil, Trash2, PlusCircle, MinusCircle } from 'lucide-react'
 import { DiceLoader } from '@/components/ui/dice-loader'
 import { Button } from '@/components/ui/button'
 import { Input } from '@/components/ui/input'
@@ -23,13 +23,14 @@ import {
   useAdminDeleteEvent,
   useAdminRooms,
 } from '@/lib/hooks/use-admin'
-import type { AdminEvent } from '@/lib/types'
+import type { AdminEvent, AdminEventSchedule } from '@/lib/types'
 
 const NONE_ROOM = '__none__'
 
-interface EventFormState {
-  title: string
-  description: string
+// ---------------------------------------------------------------------------
+// Per-schedule entry form state
+// ---------------------------------------------------------------------------
+interface ScheduleEntry {
   date: string
   startTime: string
   endTime: string
@@ -37,23 +38,186 @@ interface EventFormState {
   allDay: boolean
 }
 
-function emptyForm(): EventFormState {
-  return { title: '', description: '', date: '', startTime: '', endTime: '', roomId: NONE_ROOM, allDay: false }
+function emptySchedule(): ScheduleEntry {
+  return { date: '', startTime: '', endTime: '', roomId: NONE_ROOM, allDay: false }
 }
 
-function formFromEvent(event: AdminEvent): EventFormState {
+function scheduleFromBlock(b: AdminEventSchedule): ScheduleEntry {
   return {
-    title: event.title,
-    description: event.description ?? '',
-    date: event.date,
-    startTime: event.startTime,
-    endTime: event.endTime,
-    roomId: event.roomBlocks[0]?.roomId ?? NONE_ROOM,
-    allDay: event.allDay,
+    date: b.date,
+    startTime: b.startTime,
+    endTime: b.endTime,
+    roomId: b.roomId ?? NONE_ROOM,
+    allDay: b.allDay,
   }
 }
 
-// Dialog for create/edit
+// ---------------------------------------------------------------------------
+// Top-level event form state
+// ---------------------------------------------------------------------------
+interface EventFormState {
+  title: string
+  description: string
+  schedules: ScheduleEntry[]
+}
+
+function emptyForm(): EventFormState {
+  return { title: '', description: '', schedules: [emptySchedule()] }
+}
+
+function formFromEvent(event: AdminEvent): EventFormState {
+  const schedules =
+    event.schedules.length > 0
+      ? event.schedules.map(scheduleFromBlock)
+      : [emptySchedule()]
+  return {
+    title: event.title,
+    description: event.description ?? '',
+    schedules,
+  }
+}
+
+// ---------------------------------------------------------------------------
+// ScheduleRow — one date/room/time entry in the form
+// ---------------------------------------------------------------------------
+function ScheduleRow({
+  index,
+  entry,
+  canRemove,
+  onChange,
+  onRemove,
+  dialogId,
+}: {
+  index: number
+  entry: ScheduleEntry
+  canRemove: boolean
+  onChange: (updated: ScheduleEntry) => void
+  onRemove: () => void
+  dialogId: string
+}) {
+  const t = useTranslations('admin')
+  const tc = useTranslations('common')
+  const { data: rooms } = useAdminRooms()
+  const id = (suffix: string) => `${dialogId}-sched-${index}-${suffix}`
+
+  function field(key: keyof ScheduleEntry) {
+    return (e: React.ChangeEvent<HTMLInputElement>) =>
+      onChange({ ...entry, [key]: e.target.value })
+  }
+
+  return (
+    <div className="rounded-lg border border-border bg-background-secondary/40 p-3 space-y-3">
+      {/* Row header */}
+      <div className="flex items-center justify-between">
+        <span className="text-xs font-medium text-muted-foreground">
+          {t('events.scheduleDay', { n: index + 1 })}
+        </span>
+        {canRemove && (
+          <button
+            type="button"
+            aria-label={t('events.removeSchedule')}
+            onClick={onRemove}
+            className="text-muted-foreground hover:text-destructive transition-colors"
+          >
+            <MinusCircle className="h-4 w-4" aria-hidden="true" />
+          </button>
+        )}
+      </div>
+
+      {/* Room */}
+      <div className="space-y-1.5">
+        <Label htmlFor={id('room')} className="text-xs text-muted-foreground font-medium">
+          {t('events.room')}
+        </Label>
+        <Select
+          value={entry.roomId}
+          onValueChange={(v) => onChange({ ...entry, roomId: v })}
+        >
+          <SelectTrigger id={id('room')} className="bg-background-secondary border-border h-8 text-sm">
+            <SelectValue />
+          </SelectTrigger>
+          <SelectContent>
+            <SelectItem value={NONE_ROOM}>—</SelectItem>
+            {(rooms ?? []).map((room) => (
+              <SelectItem key={room.id} value={room.id}>{room.name}</SelectItem>
+            ))}
+          </SelectContent>
+        </Select>
+      </div>
+
+      {/* All-day toggle */}
+      <div className="flex items-start gap-2">
+        <Checkbox
+          id={id('all-day')}
+          checked={entry.allDay}
+          onCheckedChange={(checked) => onChange({
+            ...entry,
+            allDay: checked === true,
+            startTime: checked === true ? '' : entry.startTime,
+            endTime: checked === true ? '' : entry.endTime,
+          })}
+          className="mt-0.5"
+        />
+        <Label htmlFor={id('all-day')} className="text-xs text-foreground font-medium leading-tight">
+          {t('events.allDay')}
+        </Label>
+      </div>
+
+      {/* Date + optional times */}
+      <div className="grid grid-cols-1 sm:grid-cols-3 gap-2">
+        <div className="space-y-1">
+          <Label htmlFor={id('date')} className="text-xs text-muted-foreground font-medium">
+            {tc('date')}
+          </Label>
+          <Input
+            id={id('date')}
+            type="date"
+            value={entry.date}
+            onChange={field('date')}
+            required
+            className="bg-background-secondary border-border focus:border-primary/50 h-8 text-sm"
+          />
+        </div>
+        {!entry.allDay && (
+          <>
+            <div className="space-y-1">
+              <Label htmlFor={id('start')} className="text-xs text-muted-foreground font-medium">
+                {t('events.startTime')}
+              </Label>
+              <Input
+                id={id('start')}
+                type="time"
+                step={3600}
+                value={entry.startTime}
+                onChange={field('startTime')}
+                required={!entry.allDay}
+                className="bg-background-secondary border-border focus:border-primary/50 h-8 text-sm"
+              />
+            </div>
+            <div className="space-y-1">
+              <Label htmlFor={id('end')} className="text-xs text-muted-foreground font-medium">
+                {t('events.endTime')}
+              </Label>
+              <Input
+                id={id('end')}
+                type="time"
+                step={3600}
+                value={entry.endTime}
+                onChange={field('endTime')}
+                required={!entry.allDay}
+                className="bg-background-secondary border-border focus:border-primary/50 h-8 text-sm"
+              />
+            </div>
+          </>
+        )}
+      </div>
+    </div>
+  )
+}
+
+// ---------------------------------------------------------------------------
+// EventFormDialog — create / edit
+// ---------------------------------------------------------------------------
 function EventFormDialog({
   open,
   onOpenChange,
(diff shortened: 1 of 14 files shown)


lib/server/events-service.ts
```diff
@@ -287,15 +432,23 @@ export async function deleteEvent(id: string): Promise<void> {
...
+        .eq('date', block.date)
+        .lt('start_time', block.end_time)
+        .gt('end_time', block.start_time)
         .in('status', ['active', 'pending'])
 
       if (cancelError) serviceError('Internal server error', 500)
     }
   }
 
+  // Fallback: if no blocks exist, cancel by event-level date/time (legacy events)
```
Oiranca: **[Low — dead code]** This fallback block is a no-op: it builds an empty `roomIds`, then `void roomIds` / `void event` just to silence unused-var lint. Nothing is cancelled. Either implement the legacy event-level cancellation here or delete the whole `if` block (lines 444-450) — right now it only adds noise.

lib/server/events-service.ts
```diff
@@ -268,17 +415,15 @@ export async function deleteEvent(id: string): Promise<void> {
...
-    .select('room_id')
+    .select('room_id, date, start_time, end_time')
     .eq('event_id', id)
 
-  const roomIds = ((blocks ?? []) as EventRoomBlockRow[]).map((b) => b.room_id)
-
-  if (roomIds.length > 0) {
-    // Find tables in those rooms
+  // Cancel overlapping reservations for every block (multi-day aware)
+  for (const block of (blocks ?? []) as (EventRoomBlockRow & { date: string; start_time: string; end_time: string })[]) {
```
Oiranca: **[Low — N+1]** For a multi-day event this loops per block, running a `tables` select + a `reservations` update for each one. A 10-day event = 20 round trips. Consider collecting distinct `room_id`s first and cancelling per (date,time) in fewer queries, or fold the cancellation into the DB function (the create/update RPCs already cancel in-DB — delete could too, for consistency and atomicity).

lib/server/events-service.ts
```diff
@@ -142,21 +240,47 @@ export async function getEvent(id: string): Promise<AdminEvent> {
...
+      start_time: b.start_time,
+      end_time: b.end_time,
+      all_day: b.all_day,
+    }))
+
+    const admin = createSupabaseServerAdminClient()
+    const { data: result, error: rpcError } = await admin.rpc('create_event_with_blocks', {
+      p_title: title,
+      p_description: description,
+      p_blocks: blocksPayload,
```
Oiranca, line 309: **[Low — error mapping]** Any `rpcError` from the RPC collapses to a generic `500`. A block that violates the `event_room_blocks_valid_time_range` CHECK (or a bad uuid that slips past validation) is a client error, not a server fault. Inspect the PG error code and map constraint/`22P02` violations to `400`.

components/admin/events-section.tsx
```diff
@@ -316,18 +441,51 @@ function EventRow({
...
+              .filter((s) => s.roomId)
+              .map((s, i) => {
+                const roomName = (rooms ?? []).find((r) => r.id === s.roomId)?.name ?? s.roomId
+                return (
+                  <Badge key={i} variant="partial" className="text-xs">
+                    {roomName}
+                  </Badge>
+                )
+              })
+              // Deduplicate by name
```
Oiranca: **[Low — readability]** Deduplicating by mapping to `<Badge>` elements first and then `.filter`-ing on `badge.key` is hard to follow and relies on React element internals. Dedupe the data before rendering, e.g. build a unique room-name set from `schedules`, then map that to badges.

supabase/migrations/20260617000001_kim383_multi_day_events.sql
```diff
@@ -0,0 +1,259 @@
+-- KIM-383: Multi-day event support
+--
+-- Adds two new SECURITY DEFINER functions that accept a JSONB array of
+-- schedule blocks (room_id, date, start_time, end_time, all_day) so that a
+-- single event can block multiple rooms across multiple dates atomically.
+--
+-- The existing create_event_atomic / update_event_atomic single-room RPCs are
+-- intentionally left in place for backward compatibility.
+--
+-- Uses internal.is_admin() for RLS parity with the rest of the codebase.
```
Oiranca: **[Low — inaccurate comment]** This says "Uses internal.is_admin() for RLS parity", but neither function calls `is_admin()`. They are `SECURITY DEFINER`, `REVOKE`d from PUBLIC and granted only to `service_role` — admin enforcement lives in the service layer (`requireAdmin`). Reword to avoid implying an in-function admin check that does not exist.

supabase/migrations/20260617000001_kim383_multi_day_events.sql
```diff
@@ -0,0 +1,259 @@
...
+          'end_time',   v_block_rec.end_time,
+          'all_day',    v_block_rec.all_day
+        )
+      );
+
+      SELECT ARRAY(SELECT id FROM public.tables WHERE room_id = v_room_id)
+      INTO v_table_ids;
+
+      IF array_length(v_table_ids, 1) > 0 THEN
+        UPDATE public.reservations
```
Oiranca, line 106: **[Medium — silent data loss]** Creating (and updating) an event cancels overlapping member reservations (`status = 'cancelled') with no confirmation. `deleteEvent` shows a `deleteWarning` before cancelling, but create/update give the admin no heads-up. Recommend surfacing the same warning (count of reservations that will be cancelled) before the create/update mutation, so the behaviour is consistent and intentional.

supabase/migrations/20260617000001_kim383_multi_day_events.sql
```diff
@@ -0,0 +1,259 @@
...
+  -- start_time / end_time columns that the legacy model still expects).
+  SELECT
+    (blk->>'date')::date,
+    CASE WHEN (blk->>'all_day')::boolean THEN '00:00'::time
+         ELSE (blk->>'start_time')::time END,
+    CASE WHEN (blk->>'all_day')::boolean THEN '23:59'::time
+         ELSE (blk->>'end_time')::time END
+  INTO v_anchor_date, v_anchor_start, v_anchor_end
+  FROM jsonb_array_elements(p_blocks) AS blk
+  ORDER BY (blk->>'date')::date ASC,
```
Oiranca, line 65: **[Low — anchor ordering]** The anchor `ORDER BY (blk->>'start_time') ASC` sorts the raw text field, but the resolved start (the `CASE WHEN all_day THEN 00:00`) is what is stored. For an all-day block, `start_time` may be empty/absent, so an all-day block could sort before/after a timed block on the same date inconsistently with the stored `00:00`. Order by the same resolved expression you SELECT.

supabase/migrations/20260617000001_kim383_multi_day_events.sql
```diff
@@ -0,0 +1,263 @@
...
+    'room_blocks', v_blocks_json
+  );
+END;
+$$;
+
+ALTER FUNCTION "public"."create_event_with_blocks"(text, text, jsonb) OWNER TO "postgres";
+
+-- Grant same as existing event atomic functions
+REVOKE ALL ON FUNCTION "public"."create_event_with_blocks"(text, text, jsonb) FROM PUBLIC;
+GRANT ALL ON FUNCTION "public"."create_event_with_blocks"(text, text, jsonb) TO "service_role";
```
Oiranca: **[High — security]** `REVOKE ALL ... FROM PUBLIC` does not cover `anon`/`authenticated`, which Supabase has historically auto-granted EXECUTE on `public` functions (this repo needed dedicated revoke migrations for `create_event_atomic`/`update_event_atomic`). Without explicit `REVOKE EXECUTE ... FROM anon` and `FROM authenticated`, these SECURITY DEFINER mutation RPCs can become callable unauthenticated via PostgREST `/rpc/` → event creation + mass reservation cancellation. Add explicit per-role REVOKEs (and prefer `GRANT EXECUTE` over `GRANT ALL`). Same for `update_event_with_blocks` (line 263).

lib/server/events-service.ts
```diff
@@ -201,22 +333,50 @@ export async function updateEvent(
...
   const title = body.title !== undefined ? String(body.title).trim() || currentRow.title : currentRow.title
   const description =
     body.description !== undefined
       ? body.description === null
         ? null
         : String(body.description).trim() || null
       : currentRow.description
+
+  // --- Multi-block path (new) ---
+  if (Array.isArray(body.schedules) && body.schedules.length > 0) {
```
Oiranca: **[High — correctness]** Guard is `Array.isArray(body.schedules) && body.schedules.length > 0`. A PUT with `schedules: []` falls through to the legacy single-block path (`update_event_atomic`), deleting all blocks and collapsing a multi-day event to one — silently. Treat a present `schedules` array as the multi-block path; if empty, return 400. Same in `createEvent` (line 258).

lib/hooks/use-admin.ts
```diff
@@ -174,10 +174,30 @@ export function useAdminEvents() {
...
+      schedules?: EventSchedulePayload[]
+      /** Legacy single-block fields */
+      date?: string
+      startTime?: string
+      endTime?: string
+      roomId?: string | null
+      allDay?: boolean
+    }) =>
       apiClient.post<AdminEvent>(endpoints.events.list, data),
     onSuccess: () => queryClient.invalidateQueries({ queryKey: ['admin', 'events'] }),
```
Oiranca: **[Medium — stale UI]** Event mutations only invalidate `[admin, events]`. The member availability grid uses `[availability, ...]`; after an admin blocks/unblocks a room it stays stale until refetch, so members can book slots an event just blocked. Add `invalidateQueries({ queryKey: [availability] })` to all three hooks (202, 225, 233).

lib/server/events-service.ts
```diff
@@ -36,56 +36,154 @@ function resolveEventTimes(date: string, startTime: string, endTime: string, all
...
+  const roomBlocks: AdminEventRoomBlock[] = blocks.map((b) => ({
+    id: b.id,
+    roomId: b.room_id,
+    date: b.date,
+    startTime: b.start_time.slice(0, 5),
+    endTime: b.end_time.slice(0, 5),
+    allDay: b.all_day,
+  }))
+
+  const schedules: AdminEventSchedule[] = blocks.map((b) => ({
```
Oiranca: **[Medium — display bug]** `schedules` is never sorted. `toAdminEvent` (here) and `jsonToAdminEvent` (108) build it in DB/insertion order; the date-range badge reads `schedules[0]`/`schedules[last]`, so out-of-order blocks render a reversed range. Sort ascending by `(date, startTime)` in both mappers.

lib/server/events-service.ts
```diff
@@ -142,21 +240,53 @@ export async function getEvent(id: string): Promise<AdminEvent> {
...
   if (!title) serviceError('Title is required', 400)
 
-  const date = String(body.date).trim()
-  const allDay = parseAllDay(body.allDay)
-  const resolvedTimes = resolveEventTimes(date, String(body.startTime ?? '').trim(), String(body.endTime ?? '').trim(), allDay)
-
   const description = body.description ? String(body.description).trim() : null
+
+  // --- Multi-block path (new) ---
+  if (Array.isArray(body.schedules) && body.schedules.length > 0) {
```
Oiranca: **[Medium — null-room data loss]** `event_room_blocks.room_id` is NOT NULL, so a `roomId: null` schedule is silently skipped (no block); the event is created but `schedules`/`roomBlocks` return empty and edit pre-fill goes blank, losing the admins date/time input. Either require a room per schedule (400 if missing) or synthesize a schedule from the event anchor when no blocks exist. Also bound `schedules.length` (e.g. ≤366) — currently unbounded.

components/admin/events-section.tsx
```diff
@@ -98,111 +268,64 @@ function EventFormDialog({
...
-                type="date"
-                value={form.date}
-                onChange={field('date')}
-                required
-                className="bg-background-secondary border-border focus:border-primary/50"
-              />
+            <div className="space-y-3">
+              {form.schedules.map((entry, i) => (
+                <ScheduleRow
+                  key={i}
```
Oiranca: **[Low — React key]** `key={i}` on a removable list shifts indices on mid-list delete (focus/animation glitches). Give each `ScheduleEntry` a stable id (`crypto.randomUUID()`) and key on that.

lib/server/events-service.ts
Oiranca: ✅ Fixed in `831cb65` — removed the dead `if (blocks.length === 0)` no-op block entirely; the event-existence check now selects only `id`.

lib/server/events-service.ts
Oiranca: ✅ Fixed in `831cb65` — collect distinct `room_id`s once, fetch all table ids in a single `.in(...)` query into a Map, then the per-block loop only runs the reservations update. Round trips reduced from 2N to 1+N.

lib/server/events-service.ts
Oiranca: ✅ Fixed in `831cb65` — both RPC error handlers now map PG `23514`/`22P02`/`23502` to `400` (Invalid event data); other codes stay `500`.

components/admin/events-section.tsx
Oiranca: ✅ Fixed in `831cb65` — dedupe at the data level (unique roomIds from `schedules`) before rendering badges; keyed on `roomId`.

supabase/migrations/20260617000001_kim383_multi_day_events.sql
Oiranca: ✅ Fixed in `831cb65` — header comment reworded to accurately describe SECURITY DEFINER + REVOKE-from-PUBLIC + service_role-only grant with admin enforcement in the service layer.

supabase/migrations/20260617000001_kim383_multi_day_events.sql
Oiranca: ✅ Fixed in `831cb65` — both anchor `ORDER BY` clauses now use the resolved `CASE WHEN all_day THEN 00:00 ELSE start_time END` expression.

supabase/migrations/20260617000001_kim383_multi_day_events.sql
Oiranca: ℹ️ Intentionally retained for now. `deleteEvent` already surfaces `deleteWarning` before cancelling; applying the same confirmation to create/update is a separate product decision (needs a pre-count + confirm dialog). Tracked as a follow-up rather than changing behavior in this PR.

supabase/migrations/20260617000001_kim383_multi_day_events.sql
Oiranca: ✅ Fixed in `852fca0` — added explicit `REVOKE EXECUTE ... FROM anon` and `FROM authenticated` on both RPCs, switched `GRANT ALL`→`GRANT EXECUTE`, and added table-level `REVOKE ALL ON events/event_room_blocks FROM anon` as defense-in-depth.

lib/server/events-service.ts
Oiranca: ✅ Fixed in `852fca0` — multi-block path now triggers on `Array.isArray(body.schedules)`; an empty array returns `400` (and `>366` returns `400`) instead of collapsing into the legacy single-block path.

lib/hooks/use-admin.ts
Oiranca: ✅ Fixed in `852fca0` — create/update/delete event hooks now invalidate both `['admin','events']` and `['availability']` via `Promise.all`.

lib/server/events-service.ts
Oiranca: ✅ Fixed in `852fca0` — `schedules` is now sorted ascending by `(date, startTime)` in both `toAdminEvent` and `jsonToAdminEvent`, so the date-range badge is always chronological.

lib/server/events-service.ts
Oiranca: ✅ Fixed in `852fca0` — chose the synthesize approach: when an event has no room blocks, one schedule entry is reconstructed from the event anchor so edit pre-fill is no longer blank. Also added the `≤366` upper bound. (Kept room-less events supported rather than forcing a room.)

components/admin/events-section.tsx
Oiranca: ✅ Fixed in `852fca0` — `ScheduleEntry` now has a stable `id` (`crypto.randomUUID()`), used as the React `key` instead of the array index.

supabase/migrations/20260617000001_kim383_multi_day_events.sql
Oiranca: ✅ Implemented in `3265c00` (+ tests `7f89929`). Create/update now run a `previewEventConflicts` check first; if any active/pending member reservations would be cancelled, a `CancelReservationsDialog` shows the exact count and requires explicit confirmation before the mutation runs (mirrors the delete flow). New `POST /api/events/preview` endpoint + `useAdminPreviewEventConflicts` hook back it.

PR #126, opened 2026-06-19
[codex] Prevent concurrent reservation conflicts (KIM-375)

Oiranca:
## Summary

- enforce pending and active reservation overlap rules atomically with GiST exclusion constraints
- resolve existing stale and conflicting pending reservations safely before adding constraints
- map PostgreSQL exclusion violations to `SLOT_TAKEN` and expire stale pending rows before create/update
- refresh availability and show localized conflict feedback after HTTP 409 responses

## Why

Application-level overlap checks had a race window: concurrent requests could both pass validation before either write completed. The database now provides the final concurrency guard while preserving independent top/bottom bookings for removable-surface tables.

## Impact

Conflicting concurrent reservation writes return HTTP 409 instead of creating overlapping bookings. The reservation dialog reports the conflict and refreshes current availability.

## Validation

- `bash ~/.codex/scripts/validate-local.sh`
- lint passed
- typecheck passed
- tests passed: 613/613
- production build passed
- pre-push local CI passed

## Tracking

KIM-375


supabase/migrations/20260619000001_kim375_atomic_reservation_overlap_constraints.sql
```diff
@@ -0,0 +1 @@
...ed'::"public"."reservation_status" WHERE "id" IN (SELECT "id" FROM "loser"); GET DIAGNOSTICS updated_count = ROW_COUNT; EXIT WHEN updated_count = 0; END LOOP; EXECUTE 'ALTER TABLE ONLY "public"."reservations" ADD CONSTRAINT "reservations_no_pending_active_overlap_top" EXCLUDE USING "gist" ("table_id" "extensions"."gist_uuid_ops" WITH =, "tsrange"(("date" + "start_time"), ("date" + "end_time"), ''[)''::"text") WITH &&) WHERE (("status" = ANY (ARRAY[''pending''::"public"."reservation_status", ''active''::"public"."reservation_status"])) AND ("surface" IS NULL OR "surface" = ''top''::"public"."table_surface"))'; EXECUTE 'ALTER TABLE ONLY "public"."reservations" ADD CONSTRAINT "reservations_no_pending_active_overlap_bottom" EXCLUDE USING "gist" ("table_id" "extensions"."gist_uuid_ops" WITH =, "tsrange"(("date" + "start_time"), ("date" + "end_time"), ''[)''::"text") WITH &&) WHERE (("status" = ANY (ARRAY[''pending''::"public"."reservation_status", ''active''::"public"."reservation_status"])) AND ("surface" IS NULL OR "surface" = ''bottom''::"public"."table_surface"))'; EXECUTE 'ALTER TABLE "public"."reservations" DROP CONSTRAINT IF EXISTS "reservations_no_active_overlap"'; END $kim375$;
```
Oiranca: The two `ADD CONSTRAINT` statements use plain `ADD` (no `IF NOT EXISTS`) while the trailing `DROP` uses `IF EXISTS`. The migration is not re-runnable: if it fails after adding `_top` but before `_bottom` (or on a partial replay / `supabase db reset` against a partially-migrated DB), re-applying aborts with `constraint "reservations_no_pending_active_overlap_top" already exists` and blocks all later migrations. Consider `ADD CONSTRAINT IF NOT EXISTS` (or guard each with a DROP IF EXISTS first).

lib/server/reservations-service.ts
```diff
@@ -277,6 +285,21 @@ async function listActiveReservationsForConflict(input: {
   }) as ReservationRow[]
 }
 
+async function expireStalePendingReservations() {
```
Oiranca: `expireStalePendingReservations()` runs a table-wide `UPDATE` cancelling every stale pending reservation across all tables/rooms on every create AND update — it isn't scoped to the target `tableId`/`date`. It's logically safe (only touches already-expired rows) but wider than needed for the constraint backstop, adding write amplification and lock contention on `reservations` under load. Scoping the cleanup to the target table+date would suffice.

lib/server/reservations-service.ts
```diff
@@ -277,6 +285,21 @@ async function listActiveReservationsForConflict(input: {
...
+  const nowUtc = await getDatabaseNow(admin)
+  const cutoff = new Date(nowUtc.getTime() - GRACE_PERIOD_MINUTES * 60 * 1000).toISOString()
+  const { error } = await (admin.from('reservations') as unknown as StalePendingReservationsTableClient)
+    .update({ status: 'cancelled' })
+    .eq('status', 'pending')
+    .is('activated_at', null)
+    .lt('created_at', cutoff)
+
+  if (error) {
+    serviceError('Internal server error', 500)
```
Oiranca: On a failed cleanup `UPDATE`, this throws `serviceError('Internal server error', 500)`, turning a janitorial pre-expire failure into a hard 500 that aborts an otherwise-valid reservation. Since the new GiST EXCLUDE constraint already backstops real conflicts (mapped to SLOT_TAKEN via `isConflictError`/`throwSlotTaken`), a transient cleanup error (e.g. statement timeout on a large stale set) shouldn't fail the request — consider logging and continuing to the insert/update path.

supabase/migrations/20260619000001_kim375_atomic_reservation_overlap_constraints.sql
Oiranca: ✅ Fixed in `526dc82` — each `ADD CONSTRAINT` is now preceded by `DROP CONSTRAINT IF EXISTS` for the matching name (`_top`/`_bottom`) inside the `DO $kim375$` block, making the migration re-runnable after a partial failure or replay.

lib/server/reservations-service.ts
Oiranca: ✅ Fixed in `ba264f5` — `expireStalePendingReservations(tableId, date)` is now scoped with `.eq('table_id', tableId).eq('date', date)`, and both call sites pass the target table+date. The cleanup no longer touches unrelated rows; the one-time global backfill remains in the migration.

lib/server/reservations-service.ts
Oiranca: ✅ Fixed in `ac4954c` — a failed cleanup `UPDATE` now logs via `console.error` and returns normally instead of throwing a 500. Double-booking is still prevented because the new GiST EXCLUDE constraint is the authoritative backstop (`23P01` → `SLOT_TAKEN` via `isConflictError`/`throwSlotTaken`), so a swallowed cleanup error cannot produce a silent overlap.

Oiranca:
### Review fixes applied

2 QA cycles + security review passed (reservations-service suite green).

- `fix(db): make kim375 overlap-constraint migration re-runnable` — `DROP CONSTRAINT IF EXISTS` before each `ADD CONSTRAINT`.
- `perf(reservations): scope stale-pending cleanup to target table and date` — `expireStalePendingReservations(tableId, date)`; both call sites scoped.
- `fix(reservations): don't fail booking when stale-pending cleanup errors` — cleanup failure logs and continues; the GiST EXCLUDE constraint remains the authoritative double-booking backstop.

Security review: **APPROVE** — admin-client cleanup is bounded and parameterized (no injection), constraint prevents silent overlaps. One LOW non-blocking note: rescheduling across dates doesn't proactively clean the original date, but the constraint handles it independently.

supabase/migrations/20260619000001_kim375_atomic_reservation_overlap_constraints.sql
```diff
@@ -0,0 +1 @@
...ions_no_pending_active_overlap_top"'; EXECUTE 'ALTER TABLE ONLY "public"."reservations" ADD CONSTRAINT "reservations_no_pending_active_overlap_top" EXCLUDE USING "gist" ("table_id" "extensions"."gist_uuid_ops" WITH =, "tsrange"(("date" + "start_time"), ("date" + "end_time"), ''[)''::"text") WITH &&) WHERE (("status" = ANY (ARRAY[''pending''::"public"."reservation_status", ''active''::"public"."reservation_status"])) AND ("surface" IS NULL OR "surface" = ''top''::"public"."table_surface"))'; EXECUTE 'ALTER TABLE "public"."reservations" DROP CONSTRAINT IF EXISTS "reservations_no_pending_active_overlap_bottom"'; EXECUTE 'ALTER TABLE ONLY "public"."reservations" ADD CONSTRAINT "reservations_no_pending_active_overlap_bottom" EXCLUDE USING "gist" ("table_id" "extensions"."gist_uuid_ops" WITH =, "tsrange"(("date" + "start_time"), ("date" + "end_time"), ''[)''::"text") WITH &&) WHERE (("status" = ANY (ARRAY[''pending''::"public"."reservation_status", ''active''::"public"."reservation_status"])) AND ("surface" IS NULL OR "surface" = ''bottom''::"public"."table_surface"))'; EXECUTE 'ALTER TABLE "public"."reservations" DROP CONSTRAINT IF EXISTS "reservations_no_active_overlap"'; END $kim375$;
```
Oiranca, line 1: Deploy-time hazard (operational, not a correctness bug): the two `ADD CONSTRAINT ... EXCLUDE USING gist` statements build their GiST indexes non-concurrently inside the `DO` transaction, each taking an `ACCESS EXCLUSIVE` lock and a full validating scan on `reservations`. Together with the preceding backfill `LOOP`, applying this on a large production table blocks all reads/writes to `reservations` for the duration of the migration.

If the table is non-trivial at deploy time, consider mitigating — e.g. build the indexes with `CREATE INDEX CONCURRENTLY` first and attach via `ADD CONSTRAINT ... USING INDEX`, or add the constraints `NOT VALID` then `VALIDATE CONSTRAINT` in a separate step — so the exclusive lock window is minimized. Per the repo's migration policy this stays prepared-only; the user applies it, so flagging for a low-traffic deploy window if mitigation isn't added.

Oiranca: Correction to my own suggestion: the mitigations I proposed do **not** apply to `EXCLUDE` constraints. Postgres `ADD CONSTRAINT ... USING INDEX` only supports `UNIQUE`/`PRIMARY KEY`, and `NOT VALID` only supports `CHECK`/`FOREIGN KEY` — neither is accepted for a GiST `EXCLUDE` constraint, and `CREATE INDEX CONCURRENTLY` cannot run inside the migration's transaction/`DO` block. So the `ACCESS EXCLUSIVE` lock during the GiST index build is inherent to adding an EXCLUDE constraint and cannot be removed in code.

For this application the `reservations` table is small (bounded membership, 7-day booking horizon), so the build + validating scan is effectively sub-second and the lock window is negligible. The migration stays prepared-only per repo policy; recommendation is simply to apply it during a low-traffic window. No code change warranted — resolving.

PR #127, opened 2026-06-19
[codex] Trigger no-show expiry on room page load (KIM-374)

Oiranca:
## Summary

- invoke the existing `mark_no_show_reservations` server-side flow after room-page authentication
- await expiry processing before rendering the rooms UI and loading current availability
- preserve the existing authenticated cron trigger as an additive, idempotent fallback
- cover valid, missing, stale, and RPC-failure page-load paths

## Why

No-show transitions previously depended on scheduled execution. That could leave stale pending reservations visible until the next cron run. Running the existing idempotent RPC when an authenticated member enters the rooms page ensures subsequent availability requests observe current reservation state.

## Implementation

The rooms server component validates the session and active profile first, then awaits the existing server-only `markNoShowReservations()` wrapper. RPC failures remain outside the authentication catch so infrastructure errors are not misreported as expired sessions. No database function or cron behavior changed.

## Security and reliability

- the privileged RPC runs only after active-session validation
- the admin Supabase client remains server-only
- browser roles retain no direct RPC execution permission
- the database update remains idempotent and indexed

## Review

- code review: approved
- security review: approved
- testing review: approved
- quality pass 1: approved
- quality pass 2: one missing-session coverage gap found, fixed, and approved on re-review

## Validation

- focused tests: 91/91
- full tests: 608/608
- lint: passed
- typecheck: passed
- production build: passed
- pre-push local CI: passed

## Tracking

KIM-374


app/[locale]/rooms/page.tsx
```diff
@@ -18,14 +19,16 @@ export default async function RoomsPage({ params }: RoomsPageProps) {
...
   }
 
   try {
     await getCurrentUser(session)
   } catch {
-    redirect(`/${locale}/login`)
+    return redirect(`/${locale}/login`)
   }
 
+  await markNoShowReservations()
```
Comment (Oiranca) [inline app/[locale]/rooms/page.tsx]: Altitude: this triggers no-show expiry on every rooms-page render. The repo already has `app/api/cron/mark-no-show/route.ts` for exactly this, but `vercel.json` has no `crons` entry wiring it. Triggering a DB write from user navigation is a bandaid for the missing scheduled job: every load (including back/forward and concurrent members at session start) re-runs the same `UPDATE`, creating redundant write traffic and row-lock contention on `reservations` against real booking/check-in writes. Prefer wiring the existing cron into `vercel.json`. It also adds a DB round-trip to TTFB on the most-visited screen even when 0 rows expire.

Comment (Oiranca) [inline app/[locale]/rooms/page.tsx]: Robustness: `markNoShowReservations()` has no try/catch here, unlike the cron caller which wraps it. On any DB/RPC error it throws `serviceError('Internal server error', 500)`, crashing the entire rooms render — a read-only screen taken down by a background maintenance task. Consider catching and logging the failure so the page still renders (the lazy/DB-side handling already covers correctness).

Comment (Oiranca) [reply app/[locale]/rooms/page.tsx]: Triggering expiry on rooms load is the explicit requirement of KIM-374, so the on-load trigger stays. The real risk you flagged is addressed in the latest commit: `markNoShowReservations()` is now **best-effort** (wrapped in try/catch, error logged via `console.error`, render continues), so a transient DB error no longer blocks or crashes the page. Wiring the existing `app/api/cron/mark-no-show/route.ts` into a `vercel.json` `crons` schedule is a sound follow-up to remove the redundant per-load writes, but is out of scope for this ticket — noted for a separate change.

Comment (Oiranca) [reply app/[locale]/rooms/page.tsx]: ✅ Fixed — `markNoShowReservations()` is now wrapped in try/catch: on failure the error is logged server-side and the page still renders `<RoomsView />`. A background-maintenance failure can no longer take down the rooms screen. The auth-pages test was updated to expect best-effort behavior (resolves, `redirect` not called).

Oiranca:
### Review fixes applied

2 QA cycles + security review passed (613 tests green).

- `fix(rooms): make no-show expiry best-effort on page load` — `markNoShowReservations()` wrapped in try/catch; failures are logged and the rooms page still renders. A transient DB error can no longer crash the screen.
- Auth-pages test updated to expect best-effort behavior (resolves, no redirect).
- On-load trigger retained per KIM-374; moving expiry to a scheduled `vercel.json` cron noted as a follow-up.

Security review: **APPROVE** — auth gates still precede expiry, error detail stays server-side, no new authorization hole.

PR #131, opened 2026-06-19
feat(KIM-384): Saved Games — recurring table reservations with attendance tracking

Oiranca:
## Summary

- Introduces the **Saved Games** feature (KIM-384): members reserve a table for recurring play over a configurable date range (max 3 months), with automatic attendance recording, a 14-day renewal window, and event-block conflict detection enforced at both the service layer and the database via a validation trigger.
- Includes **KIM-392** (slot-relative pending expiry): pending reservations now expire 60 minutes after slot start (capped at slot end) instead of a fixed wall-clock time, implemented as a `SECURITY DEFINER` function `mark_no_show_reservations`.
- Includes **KIM-375 touch-up**: atomic reservation overlap constraint migration updated to handle surface-aware exclusion for `top`/`bottom` surfaces with a looping cancellation pass before re-adding the GIST constraints.

## Key implementation details

- **Service layer** (`lib/server/saved-games-service.ts`): `listSavedGamesForSession`, `createSavedGameForSession`, `renewSavedGameForSession`. All ownership and role checks live here — route handlers are thin.
- **API routes**: `GET/POST /api/saved-games` and `POST /api/saved-games/[id]/renew`. Both mutation routes enforce `enforceMutationSecurity` + `enforceRateLimit` before `requireAuth`.
- **Admin client only**: all writes go through `createSupabaseServerAdminClient()` (service_role). Regular reads also use admin client because the service layer enforces user scoping in the query (`user_id = session.id` for non-admins) rather than relying on RLS for server-side calls.
- **RLS** (migration 12): `saved_games` and `saved_game_attendances` have RLS enabled with SELECT-only policies for authenticated users (own rows or admin). `GRANT ALL` to `service_role` only. All trigger functions revoked from `PUBLIC`, `anon`, and `authenticated`.
- **i18n**: `en.json` and `es.json` updated with full key parity (verified — 0 missing keys on either side).

## Migrations (files only — NOT executed)

Apply manually with `supabase db push` after reviewing locally:

| File | Purpose |
|------|---------|
| `20260619000001` | KIM-375: atomic overlap constraints for top/bottom surfaces |
| `20260619000002` | KIM-392: slot-relative `mark_no_show_reservations` function |
| `20260619000003` | KIM-384: `saved_games` table + validation trigger function |
| `20260619000004` | KIM-384: `saved_game_attendances` table |
| `20260619000005` | KIM-384: indexes on both tables |
| `20260619000006–07` | KIM-384: validation trigger (attach) |
| `20260619000008–09` | KIM-384: attendance counter trigger (attach) |
| `20260619000010–11` | KIM-384: event cancellation trigger (attach) |
| `20260619000012` | KIM-384: RLS policies + grants + function revokes |
| `20260619000013–14` | KIM-384: bottom-reservation conflict guard trigger (attach) |
| `20260619000015` | KIM-384: revoke reservation trigger function from public roles |
| `20260619000016` | KIM-384: allow event cancellation to update saved game status |
| `20260619000017–19` | KIM-384: atomic attendance-on-activation trigger + execute revocation |

## Review fixes

- Removed the global database update from the GET/list path; expired status is derived without mutating state.
- Made attendance recording atomic with reservation activation through a database trigger.
- Invalidated availability caches after renewal.
- Recomputed Saved Game end dates when switching modes.

## Test plan

- [x] 639/639 unit tests pass (Vitest)
- [x] E2E runners pass
- [x] TypeScript typecheck passes (`next typegen && tsc --noEmit`)
- [x] ESLint: no warnings or errors
- [x] i18n parity verified (en.json == es.json key set)
- [x] `.env.local` NOT committed (confirmed git-ignored)
- [x] No hardcoded secrets or credentials in any changed file
- [x] Migrations committed as SQL files only — no `supabase db push` executed
- [ ] Manual: run `supabase db reset` locally to validate migration sequence
- [ ] Manual: run `supabase db push` to apply to target environment


lib/server/saved-games-service.ts
```diff
@@ -0,0 +1,200 @@
...
+  const today = getCurrentClubDate()
+  if (startDate < today) serviceError('SAVED_GAME_START_IN_PAST', 400)
+  if (endDate < startDate) serviceError('SAVED_GAME_INVALID_RANGE', 400)
+  if (endDate > getMaxEndDate(startDate)) serviceError('SAVED_GAME_MAX_DURATION', 400)
+}
+
+export async function listSavedGamesForSession(session: SessionUser): Promise<SavedGame[]> {
+  const admin = createSupabaseServerAdminClient()
+  const today = getCurrentClubDate()
+  const { error: completionError } = await admin
```
Oiranca: This GET-backed service performs a global admin-client update before applying caller scope. That turns a safe read into a CSRF-reachable write and lets any authenticated read mutate unrelated users’ rows. Please derive the effective completed status on read, or move persistence to a protected background/write path.

lib/hooks/use-reservations.ts
```diff
@@ -92,3 +92,31 @@ export function useCancelReservation() {
...
+      queryClient.invalidateQueries({ queryKey: ['availability'] }),
+    ]),
+  })
+}
+
+export function useRenewSavedGame() {
+  const queryClient = useQueryClient()
+  return useMutation({
+    mutationFn: (id: string) => apiClient.post<SavedGame>(endpoints.savedGames.renew(id), {}),
+    onSuccess: () => queryClient.invalidateQueries({ queryKey: ['saved-games'] }),
```
Oiranca: Renewal creates a new active saved-game period and changes future bottom-surface availability, but this only invalidates the saved-games cache. Please invalidate the availability queries too.

lib/server/reservations-service.ts
```diff
@@ -1063,5 +1100,7 @@ export async function activateReservationByTable(
     serviceError('CHECK_IN_ALREADY_ACTIVE', 409)
   }
 
+  await recordSavedGameAttendance(updated as ReservationRow)
```
Oiranca: Attendance is written after the reservation has already been committed as active. If this insert fails, the API returns 500 although check-in succeeded; retry then returns CHECK_IN_ALREADY_ACTIVE and attendance is never recorded. Please make activation and attendance atomic, preferably with a database trigger in the same transaction.

components/rooms/reservation-dialog.tsx
```diff
@@ -194,6 +227,33 @@ export function ReservationDialog({ table, open, onClose }: ReservationDialogPro
...
+          {table.type === 'removable_top' && (
+            <div className="grid grid-cols-2 gap-2" role="radiogroup" aria-label={t('reservationType')}>
+              {(['standard', 'saved_game'] as const).map((mode) => (
+                <button
+                  key={mode}
+                  type="button"
+                  role="radio"
+                  aria-checked={reservationMode === mode}
+                  onClick={() => {
+                    setReservationMode(mode)
```
Oiranca, line 239: When the user changes the date in standard mode and later switches back to Saved Game, savedGameEndDate can retain the previous date range. Recompute the maximum end date when entering saved-game mode so the form cannot submit a stale/invalid range.

Oiranca (commented):
Focused review: correctness, security, and testability findings below.

lib/server/saved-games-service.ts
Oiranca: ✅ Fixed in f2e65c5. The list path no longer writes globally; expired status is derived in memory and covered by a regression test.

lib/hooks/use-reservations.ts
Oiranca: ✅ Fixed in f2e65c5. Saved-game renewal now invalidates both saved-games and availability query caches.

lib/server/reservations-service.ts
Oiranca: ✅ Fixed in f2e65c5. Attendance recording moved to a secured database trigger, so it commits atomically with the reservation activation update.

components/rooms/reservation-dialog.tsx
Oiranca: ✅ Fixed in f2e65c5. Entering Saved Game mode now recomputes the maximum end date from the currently selected start date.

PR #133, opened 2026-06-19
persist reservation E2E runners into repo (qa/e2e/)

Oiranca:
## Summary

- Adds 4 standalone Playwright/Node.js E2E runners to `qa/e2e/` (previously ephemeral in `/tmp/alea-qa`)
- Parameterizes all hardcoded paths: dedicated `.env.e2e.local` resolved relative to runner file, optional `CHROME_PATH` env override, `E2E_BASE_URL` env override
- Adds `qa/e2e/README.md` documenting prerequisites, setup, and what each runner validates
- Adds `qa/e2e/package.json` pinning `playwright` + `dotenv` as devDependencies
- Appends `qa/e2e/*.png`, `qa/e2e/*.webm`, `qa/e2e/*.pdf` artifact globs to `.gitignore`

## Review fixes

- Added an explicit `E2E_ALLOW_DESTRUCTIVE=1` safety gate and dedicated `.env.e2e.local`.
- Defaulted to Playwright bundled Chromium for Linux/CI portability.
- Restricted cleanup to exact fixture IDs.
- Made skipped checks distinct from failures.
- Seeded equipment into the room pool and removed false-positive assertions.

## Runners

| File | What it validates |
|---|---|
| `qa-reservation-lifecycle.mjs` | Create → slot blocked → check-in → `status: active` → second check-in → 409 `CHECK_IN_ALREADY_ACTIVE` |
| `qa-reservation-cancellation.mjs` | Cancel within cutoff → 200 + slot released; cancel within 60 min → 403 `CANCELLATION_CUTOFF` (member role) |
| `qa-no-show-expiry.mjs` | Cron auth guards (401 without/wrong secret); backdated pending → cron → `no_show` transition |
| `qa-reservation-equipment.mjs` | Equipment booking → availability blocked → unknown equipment 400 `INVALID_ROOM_EQUIPMENT` → conflict 409 `EQUIPMENT_ALREADY_RESERVED` |

## Validation results

**`node --check` (syntax):** All 4 files pass.

**Vitest isolation:** `vitest.config.mts` include glob is `__tests__/**/*.{test,spec}.{ts,tsx}` — `qa/e2e/` is not matched.

**`pnpm build`:** Passes with no errors or warnings.

**Run results from repo path (`node qa/e2e/<runner>.mjs`):**

| Runner | Result | Notes |
|---|---|---|
| `qa-no-show-expiry.mjs` | **8/8 passed** | Clean pass |
| `qa-reservation-lifecycle.mjs` | BLOCKED (env) | See below |
| `qa-reservation-cancellation.mjs` | BLOCKED (env) | See below |
| `qa-reservation-equipment.mjs` | BLOCKED (env) | See below |

**Environment regression (pre-existing, not caused by this PR):**

Three runners remain blocked on the current `develop` base at `POST /api/reservations` because `get_database_time` is correctly revoked from `authenticated` while the service still uses the session client there. PR #131 already fixes this by using the admin client; do not restore authenticated execute permission. Re-run the live runners after #131 is merged or rebased into this branch.

## How to run

```bash
pnpm exec next dev -p 3001
# ensure .env.e2e.local has E2E_ALLOW_DESTRUCTIVE=1 and PLAYWRIGHT_QA_USER/PASSWORD, PLAYWRIGHT_QA_SECONDARY_USER/PASSWORD,
# NEXT_PUBLIC_SUPABASE_URL, SUPABASE_SECRET_DEFAULT_KEY, CRON_SECRET
cd qa/e2e && npm install && npx playwright install chromium && cd ../..
node qa/e2e/qa-no-show-expiry.mjs
node qa/e2e/qa-reservation-lifecycle.mjs
node qa/e2e/qa-reservation-cancellation.mjs
node qa/e2e/qa-reservation-equipment.mjs
```

### DIFF
--- a/qa/e2e/qa-reservation-equipment.mjs
+++ b/qa/e2e/qa-reservation-equipment.mjs
@@ -8,19 +8,14 @@
  *  4. Use unknown equipment ID → assert 400 INVALID_ROOM_EQUIPMENT
  */
 import { chromium } from 'playwright';
-import dotenv from 'dotenv';
-import { fileURLToPath } from 'node:url';
-import path from 'node:path';
+import { chromiumLaunchOptions, env, requireE2EEnv } from './env.mjs';
 
-const envPath = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../../.env.local');
-dotenv.config({ path: envPath });
-const env = process.env;
 const required = [
   'PLAYWRIGHT_QA_USER', 'PLAYWRIGHT_QA_PASSWORD',
   'PLAYWRIGHT_QA_SECONDARY_USER', 'PLAYWRIGHT_QA_SECONDARY_PASSWORD',
   'NEXT_PUBLIC_SUPABASE_URL', 'SUPABASE_SECRET_DEFAULT_KEY',
 ];
-for (const name of required) if (!env[name]) throw new Error(`Missing env var: ${name}`);
+requireE2EEnv(required);
 
 const appUrl = process.env.E2E_BASE_URL || 'http://localhost:3001';
 const supabaseUrl = env.NEXT_PUBLIC_SUPABASE_URL;
@@ -36,12 +31,9 @@ const check = (name, pass, evidence) => {
   if (!pass) throw new Error(`FAIL [${name}]: ${JSON.stringify(evidence)}`);
 };
 
-const created = { equipmentId: null, reservation1Id: null, reservation2Id: null, extraTableId: null };
+const created = { equipmentId: null, reservation1Id: null, reservation2Id: null, extraTableId: null, roomDefaultSeeded: false };
 
-const browser = await chromium.launch({
-  headless: true,
-  executablePath: process.env.CHROME_PATH || '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome',
-});
+const browser = await chromium.launch(chromiumLaunchOptions());
 
 try {
   // ── 1. Create a fresh equipment item via admin REST ───────────────────────
@@ -84,6 +76,13 @@ try {
   } else {
     check('fixture table 2 (same room)', Boolean(table2.id), { table2 });
   }
+
+  const roomDefaultResp = await rest('room_default_equipment', {
+    method: 'POST',
+    body: JSON.stringify({ room_id: table.room_id, equipment_id: created.equipmentId }),
+  });
+  created.roomDefaultSeeded = roomDefaultResp.status === 201;
+  check('equipment assigned to fixture room', created.roomDefaultSeeded, { status: roomDefaultResp.status });
 
   // ── 4. Compute date: tomorrow ─────────────────────────────────────────────
   const timeResp = await rest('rpc/get_database_time', { method: 'POST', body: '{}' });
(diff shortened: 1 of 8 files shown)


Oiranca (commented):
Focused review: safety, portability, and E2E reliability findings below.

qa/e2e/qa-reservation-lifecycle.mjs
```diff
@@ -0,0 +1,222 @@
...
+ *  3. activate/check-in
+ *  4. assert status active + attendance recorded
+ *  5. second activate rejected (CHECK_IN_ALREADY_ACTIVE)
+ */
+import { chromium } from 'playwright';
+import dotenv from 'dotenv';
+import { fileURLToPath } from 'node:url';
+import path from 'node:path';
+
+const envPath = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../../.env.local');
```
Oiranca: These runners load the main app environment and immediately perform service-role writes/deletes against its Supabase target. Please use a dedicated E2E env file and require an explicit destructive-test opt-in before any privileged request.

qa/e2e/qa-reservation-lifecycle.mjs
```diff
@@ -0,0 +1,222 @@
...
+const check = (name, pass, evidence) => {
+  checks.push({ name, pass, evidence });
+  if (!pass) throw new Error(`FAIL [${name}]: ${JSON.stringify(evidence)}`);
+};
+
+const created = { reservationId: null, tableId: null };
+
+const browser = await chromium.launch({
+  headless: true,
+  executablePath: process.env.CHROME_PATH || '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome',
```
Oiranca: This macOS-only Chrome default breaks Linux/CI even after Playwright installs Chromium. Let Playwright use its bundled browser by default and apply executablePath only when CHROME_PATH is explicitly set.

qa/e2e/qa-reservation-lifecycle.mjs
```diff
@@ -0,0 +1,222 @@
...
+      check('second check-in rejected (409)', secondResp.status() === 409, {
+        status: secondResp.status(),
+        message: secondBody?.message,
+      });
+      check('second check-in message is CHECK_IN_ALREADY_ACTIVE', secondBody?.message === 'CHECK_IN_ALREADY_ACTIVE', {
+        message: secondBody?.message,
+      });
+    } else {
+      // Log as non-blocking; timing might make fixture impossible
+      checks.push({ name: 'check-in fixture', pass: false, evidence: { reason: 'Could not insert today fixture' } });
```
Oiranca: This documented skip path records pass: false, so the final passed < total check makes the runner fail by time of day. Track skipped checks separately from failures.

qa/e2e/qa-reservation-lifecycle.mjs
```diff
@@ -0,0 +1,222 @@
...
+  const profileResp2 = await rest(
+    `profiles?select=id&member_number=eq.${encodeURIComponent(env.PLAYWRIGHT_QA_USER)}&limit=1`
+  ).catch(() => null);
+  if (profileResp2) {
+    const [p2] = await json(profileResp2).catch(() => []);
+    if (p2?.id && created.tableId) {
+      const tomorrow2 = (() => {
+        const d = new Date(); d.setUTCDate(d.getUTCDate() + 1); return d.toISOString().slice(0, 10);
+      })();
+      await rest(
```
Oiranca: This cleanup can delete unrelated QA-user reservations because it filters by user/table/date instead of fixture identity. Track the injected check-in reservation ID and delete only exact IDs created by this run.

qa/e2e/qa-reservation-equipment.mjs
```diff
@@ -0,0 +1,220 @@
...
+  const eqAvailResp = await page1.request.get(
+    `${appUrl}/api/rooms/${table.room_id}/available-equipment?date=${date}&startTime=${startTime}&endTime=${endTime}`
+  );
+  const eqAvailBody = await json(eqAvailResp);
+  if (eqAvailResp.status() === 200 && Array.isArray(eqAvailBody)) {
+    const eqEntry = eqAvailBody.find((e) => e.id === created.equipmentId);
+    if (eqEntry) {
+      check('equipment shows as unavailable after booking', eqEntry.available === false, { eqEntry });
+    } else {
+      checks.push({ name: 'equipment availability', pass: true, evidence: { note: 'not in pool (may be locked to another room)' } });
```
Oiranca: This branch reports success when the equipment is absent, so the runner can pass without testing availability at all. Seed the equipment into the room-default pool, assert the endpoint response, and fail if the item is missing.

qa/e2e/qa-reservation-lifecycle.mjs
Oiranca: ✅ Fixed in 52a6741. Runners now load a dedicated `.env.e2e.local` and refuse privileged writes unless `E2E_ALLOW_DESTRUCTIVE=1` is set.

qa/e2e/qa-reservation-lifecycle.mjs
Oiranca: ✅ Fixed in 52a6741. Playwright now uses bundled Chromium by default and only applies `CHROME_PATH` when explicitly configured.

qa/e2e/qa-reservation-lifecycle.mjs
Oiranca: ✅ Fixed in 52a6741. Skipped fixtures are tracked separately and no longer count as failures.

qa/e2e/qa-reservation-lifecycle.mjs
Oiranca: ✅ Fixed in 52a6741. The injected check-in fixture ID is retained and cleanup deletes only exact IDs created by the run.

qa/e2e/qa-reservation-equipment.mjs
Oiranca: ✅ Fixed in 52a6741. The runner seeds the equipment into the room pool and now fails unless the endpoint returns that item as unavailable.

PR #136, opened 2026-06-24
fix(security): replace xlsx with exceljs (2 HIGH CVEs) [KIM-395]

Oiranca:
## Summary

Removes the `xlsx` (SheetJS) dependency, which carries two **HIGH** advisories with no patched release on npm, and replaces the XLSX member-import parsing path with **`exceljs`** (independent implementation, not affected).

| Advisory | Issue |
| --- | --- |
| GHSA-4r6h-8v6p-xvw6 | Prototype pollution |
| GHSA-5pgg-2g8v-p4x9 | ReDoS |

`node-xlsx` was rejected because it wraps the same vulnerable SheetJS core.

## Changes

- `package.json` / `pnpm-lock.yaml`: drop `xlsx`, add `exceljs@^4.4.0` (ships its own types).
- `lib/server/users-service.ts`: `extractSpreadsheetCsv` now loads the workbook via `exceljs` and is `async`; `normalizeMemberImportSource` becomes `async` (its only caller, `importMembersFromSource`, already awaited it). **CSV and ODT paths are untouched.** Output row shape preserved.
- `__tests__/server/member-import.test.ts`: fixtures rebuilt with `exceljs`, async calls awaited, error cases switched to `rejects.toThrow`. Added parity tests for numeric member numbers and ragged rows; ODT cases restored.

## Scope note

Parsing is intentionally **not** extracted into its own module here — that is tracked separately in **KIM-407**. This PR is limited to the security-relevant library swap.

## Validation

- `pnpm typecheck` — pass
- `pnpm test` — **669/669** pass, **0 skipped**
- `pnpm build` — pass
- `pnpm audit --prod --audit-level high` — both `xlsx` advisories gone

Refs KIM-395

### DIFF
--- a/lib/server/users-service.ts
+++ b/lib/server/users-service.ts
@@ -250,7 +250,7 @@ function assertSourceArchiveMatchesExtension(extension: 'xlsx' | 'odt', bytes: U
   }
 }
 
-export async function extractSpreadsheetCsv(bytes: Uint8Array): Promise<string> {
+async function extractSpreadsheetCsv(bytes: Uint8Array): Promise<string> {
   const wb = new ExcelJS.Workbook()
 
   try {
--- a/lib/server/users-service.ts
+++ b/lib/server/users-service.ts
@@ -250,7 +250,7 @@ function assertSourceArchiveMatchesExtension(extension: 'xlsx' | 'odt', bytes: U
   }
 }
 
-async function extractSpreadsheetCsv(bytes: Uint8Array): Promise<string> {
+export async function extractSpreadsheetCsv(bytes: Uint8Array): Promise<string> {
   const wb = new ExcelJS.Workbook()
 
   try {
--- a/lib/server/users-service.ts
+++ b/lib/server/users-service.ts
@@ -4,7 +4,7 @@ import { createSupabaseServerAdminClient } from '@/lib/supabase/server'
 import { serviceError } from '@/lib/server/service-error'
 import type { TablesUpdate } from '@/lib/supabase/types'
 import { memberNumberSchema } from '@/lib/validations/auth'
-import { read, utils } from 'xlsx'
+import ExcelJS from 'exceljs'
 import { type PublicProfileRow, toPublicUser } from '@/lib/server/profile-mappers'
 
 type ProfilesQuery = {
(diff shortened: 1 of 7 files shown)


lib/server/users-service.ts
```diff
@@ -250,31 +250,38 @@ function assertSourceArchiveMatchesExtension(extension: 'xlsx' | 'odt', bytes: U
...
 }
 
-function extractSpreadsheetCsv(bytes: Uint8Array) {
-  let workbook: ReturnType<typeof read>
+async function extractSpreadsheetCsv(bytes: Uint8Array): Promise<string> {
+  const wb = new ExcelJS.Workbook()
 
   try {
-    workbook = read(bytes, { type: 'array', cellText: false, cellDates: false })
+    await wb.xlsx.load(bytes.buffer as ArrayBuffer)
```
Oiranca: `bytes.buffer` can include bytes outside this `Uint8Array` view when the upload buffer is a slice. ExcelJS will parse the whole backing `ArrayBuffer`, so a valid sliced upload can fail because unrelated prefix/suffix bytes are included, and tests miss this because they always pass a zero-offset array. Pass an exact buffer instead, e.g. `bytes.buffer.slice(bytes.byteOffset, bytes.byteOffset + bytes.byteLength)`, and add a regression test with `new Uint8Array(padded.buffer, offset, length)`.

Oiranca: ✅ Fixed in `4c2f8a7`. `wb.xlsx.load(bytes.buffer)` now passes the exact byte range: `bytes.buffer.slice(bytes.byteOffset, bytes.byteOffset + bytes.byteLength)`, so ExcelJS no longer parses unrelated prefix/suffix bytes when the `Uint8Array` is a sliced view. Added a genuine regression test (in `users-service.test.ts`, calling `extractSpreadsheetCsv` directly with a non-zero `byteOffset` view + a crafted trailing fake zip EOCD record) — it fails on the old code and passes with the fix. Note: the test was intentionally placed in `users-service.test.ts` (next to the fixed code) rather than `member-import.test.ts` to avoid a merge conflict with #146 on develop.

lib/server/users-service.ts
```diff
@@ -250,31 +250,38 @@ function assertSourceArchiveMatchesExtension(extension: 'xlsx' | 'odt', bytes: U
   }
 }
 
-function extractSpreadsheetCsv(bytes: Uint8Array) {
-  let workbook: ReturnType<typeof read>
+export async function extractSpreadsheetCsv(bytes: Uint8Array): Promise<string> {
```
Oiranca: This helper was made `export` only so the regression test can call it directly. That widens the production module surface for a private parsing detail and makes future refactors harder. Please keep `extractSpreadsheetCsv` private and cover this through `normalizeMemberImportSource`/`importMembersFromSource` instead; the conflict with #146 can still be avoided by moving the test file, not by exporting internals.

Oiranca (commented):
Review pass: one new testability/API-surface issue found after the sliced-buffer fix.

lib/server/users-service.ts
Oiranca: ✅ Fixed in 2ce0739e11ec97fc76bc9f4c81fea3382bb0c78d.

`extractSpreadsheetCsv` is module-private again (removed the `export`). The regression test (byte-offset-sliced backing buffer with a crafted trailing zero-record EOCD) now goes through `normalizeMemberImportSource` instead — it stayed in `__tests__/server/users-service.test.ts` (not `member-import.test.ts`) so the #146 conflict is still avoided by test-file placement, not by exporting an internal helper. Ran the full users-service and member-import test suites plus typecheck locally; all green.

PR #138, opened 2026-06-24
feat(security): defense-in-depth guard for member-scoped reads [KIM-397]

Oiranca:
## Summary

`listVisibleReservations` and `listSavedGamesForSession` read member data through the **admin client** (RLS bypass) and previously relied solely on a single conditional `.eq('user_id', session.id)` filter for isolation. If that filter ever regresses — and after RLS is removed in migration Phase 2 — there is no second line of defense.

This PR adds a structural, defense-in-depth guard.

## Changes

- **`lib/server/data-scoping.ts`** (new): `assertMemberRowsScoped<T extends { user_id: string }>(rows, session)` — for a non-admin session, throws `500` if any returned row's `user_id` differs from `session.id`; admins pass through; returns rows otherwise. Independent of the query builder.
- **`reservations-service.ts` / `saved-games-service.ts`**: call the guard on raw rows **before** mapping to the public shape. Existing `.eq('user_id', session.id)` filters are kept — the guard is additive, not a replacement.
- **`CLAUDE.md`**: documents the invariant in *Key conventions*.

Switching these reads to the session client is intentionally **out of scope** (tracked in KIM-405).

## Tests

- `data-scoping.test.ts` (new): member own-rows pass; member + foreign row throws 500 with exact message; admin exempt; empty rows OK.
- Isolation cases added to reservations/saved-games service tests proving a member session throws rather than receiving another member's rows.

## Validation

- `pnpm typecheck` — pass
- `pnpm test` — **681/681** pass (+12), 0 skipped
- `pnpm build` — pass

Closes KIM-397

### DIFF
--- a/__tests__/server/reservations-service.test.ts
+++ b/__tests__/server/reservations-service.test.ts
@@ -84,6 +84,11 @@ const eventRoomBlocksState: EventRoomBlockRow[] = []
 let reservationInsertError: { code: string } | null = null
 let reservationUpdateError: { code: string } | null = null
 let sessionDatabaseTimeDenied = false
+// Regression-test knob: simulates the `.eq('user_id', ...)` query filter being
+// accidentally removed/bypassed at the query layer, so the mock returns mixed
+// rows across users. Used to prove assertMemberRowsScoped() itself throws,
+// independent of the query filter working correctly.
+let bypassUserIdFilterInMock = false
 
 function createDatabaseTimeRpc() {
   return vi.fn(async (fn: string) => {
@@ -223,6 +228,12 @@ function buildSelectChain<T>(rows: T[], hydrate?: (row: T) => T) {
 
   return {
     eq(column: string, value: string) {
+      // Regression-test hook: when bypassUserIdFilterInMock is set, pretend the
+      // real query builder dropped the `.eq('user_id', ...)` filter so foreign
+      // rows leak through, exercising the assertMemberRowsScoped() guard.
+      if (column === 'user_id' && bypassUserIdFilterInMock) {
+        return this
+      }
       current = current.filter((row) => String((row as Record<string, unknown>)[column]) === value)
       return this
     },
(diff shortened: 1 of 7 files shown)


Oiranca (commented):
Review pass: defense-in-depth implementation is good, but one service-level test does not prove the invariant it claims to cover.

__tests__/server/reservations-service.test.ts
```diff
@@ -786,6 +786,50 @@ describe('reservations service', () => {
...
+      })
+      const t1 = tablesState.get('t1')!
+      reservationsState.push({
+        ...foreignReservation,
+        profiles: profilesMap.get('999') ?? null,
+        tables: { name: t1.name, rooms: roomsMap.get(t1.room_id) ?? null },
+      })
+
+      // Member session queries all reservations (no user filter passed in)
+      // Even if the query somehow returned mixed rows, the defense-in-depth guard
```
Oiranca, line 821: This test comment says the defense-in-depth guard should reject mixed/foreign rows, but the test only proves the normal query filter excludes the foreign row. If the `eq('user_id', session.id)` filter were accidentally removed, this mock would need to return mixed rows and assert a 500 from `assertMemberRowsScoped()`. Please add a regression test that forces the query layer to leak a foreign reservation and verifies the guard throws before mapping.

Oiranca: Good catch — added a genuine regression test in commit c00b52e: a new `it` case sets a mock-only toggle (`bypassUserIdFilterInMock`) that makes the mocked query builder ignore the `.eq('user_id', ...)` filter, so `reservationsState` returns a mixed foreign row (`user_id: '999'`) alongside the member's own rows — simulating the filter being accidentally removed at the query layer.

The test then asserts `listVisibleReservations({ session: memberSession })` rejects with `{ name: 'ServiceError', statusCode: 500, message: 'Data isolation violation: member read returned foreign rows' }`, proving `assertMemberRowsScoped()` itself catches the leak before any row mapping happens.

As a sanity check, I temporarily commented out the `assertMemberRowsScoped(...)` call in `lib/server/reservations-service.ts` and re-ran the suite — the new test failed (the foreign row leaked through into the mapped result), confirming it actually exercises the guard rather than just the query filter. Guard restored, all 89 tests in the file pass.

PR #141, opened 2026-06-24
feat(security): rate-limit store, edge crypto split, runtime cookie flag [KIM-401/402/403]

Oiranca:
## Summary

Three related security-layer hardening fixes, all centered on `lib/server/security.ts`. They share the same file, so they are delivered together on one branch with one PR (separate branches would conflict on every merge).

### KIM-401 — Shared rate-limit store (Vercel blocker)
The rate limiter used a `globalThis` Map. On serverless, each instance has its own store, so the limit is trivially bypassed by rotating instances (affects login/activate/register). 

- `enforceRateLimit(...)` is now **async**.
- When `UPSTASH_REDIS_REST_URL` + `UPSTASH_REDIS_REST_TOKEN` are set → `@upstash/ratelimit` sliding window backed by Upstash Redis, honoring the existing `RATE_LIMIT_POLICIES`. The Redis client is a memoized singleton and `Ratelimit` instances are cached per bucket.
- Otherwise → the original in-memory store (so local dev / tests / builds work with no Redis), with a one-time production warning.
- All route-handler call sites updated to `await`.

### KIM-402 — Isolate Node crypto from the Edge/middleware path (Vercel blocker)
`security.ts` statically imported Node `crypto` (`timingSafeEqual`, `createHash`), and `middleware.ts` (Edge Runtime) imported from it → Node crypto pulled into the Edge bundle.

- New `lib/server/security-edge.ts` holds the Edge-safe helpers (`isSecureContext`, `createCsrfToken`, `getCsrfCookieOptions`, `getSupabaseCookieOptions`, `ensureCsrfCookie`, CSRF name constants) — Web Crypto only.
- `security.ts` keeps the Node-only pieces (`tokensMatch`, `enforceMutationSecurity`, rate limiter) and **re-exports** the edge-safe helpers, so existing `@/lib/server/security` imports are unchanged.
- `middleware.ts` imports from `security-edge`. **Build no longer emits the Edge Runtime crypto warning.**

### KIM-403 — Cookie Secure flag at runtime
`isSecureContext()` previously read `NEXT_PUBLIC_APP_URL`, which Next.js inlines at build time, so one artifact would be wrong on either HTTP staging or HTTPS prod. Now it reads `COOKIE_SECURE` at runtime (falling back to `NODE_ENV === 'production'`). Documented in `.env.example`.

## Validation

- `pnpm typecheck` — pass
- `pnpm lint` — pass
- `pnpm test` — **671/671** pass, 0 failures, 0 new skips (rate-limit tests awaited; cookie-secure tests rewritten for the runtime logic)
- `pnpm build` — pass, **no Edge Runtime crypto warning**

## Follow-up for the user
To activate the shared rate-limit store in production, provision an Upstash Redis instance and set `UPSTASH_REDIS_REST_URL` / `UPSTASH_REDIS_REST_TOKEN` in Vercel. Until then the limiter runs in-memory (per-instance) and logs a warning.

Closes KIM-401
Closes KIM-402
Closes KIM-403

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -21,10 +21,15 @@ SUPABASE_SECRET_DEFAULT_KEY=sb_secret_xxxxxxxxxxxxxxxx
 # APPLICATION
 # ============================================================
 
-# Base URL of the app (used for auth callbacks, redirects, and server-side cookie security decisions)
-# IMPORTANT: Supabase and CSRF cookies are only marked `Secure` when this is set to an `https://` URL.
-# Use `http://localhost:3000` for local development, but set this to your real `https://` app URL in deployed environments.
+# Base URL of the app (used for auth callbacks and redirects).
+# Previously also drove cookie Secure flag — that role is now taken by COOKIE_SECURE below.
 NEXT_PUBLIC_APP_URL=http://localhost:3000
+
+# Controls the `Secure` flag on auth and CSRF cookies. Evaluated at RUNTIME (not build time),
+# so the same build artifact works correctly on both HTTP staging and HTTPS production.
+# Set to `true` on any deployment that is served over HTTPS.
+# If unset, defaults to `true` when NODE_ENV=production and `false` otherwise.
+COOKIE_SECURE=false
 
 # Enable this only when the deployment ingress strips and rewrites both `x-real-ip`
 # and `x-forwarded-for` before requests reach the app runtime.
@@ -48,6 +53,18 @@ NEXT_PUBLIC_ASSOCIATION_URL=
 # Club timezone override (IANA name, e.g. Atlantic/Canary).
 # Defaults to the server's system timezone if not set.
 # CLUB_TIMEZONE=
+
+# ============================================================
+# RATE LIMITING (optional — shared store via Upstash Redis)
+# ============================================================
+
+# When both variables are set, the rate limiter uses Upstash Redis (sliding-window,
+# shared across all serverless instances). Without them, an in-memory Map is used
+# (per-instance only — fine for local dev and tests, but bypassable in production
+# by rotating instances).
+# Dashboard: console.upstash.com → your Redis database → REST API
+UPSTASH_REDIS_REST_URL=
+UPSTASH_REDIS_REST_TOKEN=
 
 # ============================================================
 # CRON
--- a/.env.example
+++ b/.env.example
@@ -28,8 +28,10 @@ NEXT_PUBLIC_APP_URL=http://localhost:3000
 # Controls the `Secure` flag on auth and CSRF cookies. Evaluated at RUNTIME (not build time),
 # so the same build artifact works correctly on both HTTP staging and HTTPS production.
 # Set to `true` on any deployment that is served over HTTPS.
-# If unset, defaults to `true` when NODE_ENV=production and `false` otherwise.
-COOKIE_SECURE=false
+# If unset, defaults to `true` when NODE_ENV=production and `false` otherwise — leave this
+# unset in production so a deployed HTTPS environment never accidentally ships insecure cookies.
+# For local HTTP dev only, uncomment the next line:
+# COOKIE_SECURE=false
 
 # Enable this only when the deployment ingress strips and rewrites both `x-real-ip`
 # and `x-forwarded-for` before requests reach the app runtime.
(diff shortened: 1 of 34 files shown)


.env.example
```diff
@@ -21,11 +21,16 @@ SUPABASE_SECRET_DEFAULT_KEY=sb_secret_xxxxxxxxxxxxxxxx
...
-# Use `http://localhost:3000` for local development, but set this to your real `https://` app URL in deployed environments.
+# Base URL of the app (used for auth callbacks and redirects).
+# Previously also drove cookie Secure flag — that role is now taken by COOKIE_SECURE below.
 NEXT_PUBLIC_APP_URL=http://localhost:3000
 
+# Controls the `Secure` flag on auth and CSRF cookies. Evaluated at RUNTIME (not build time),
+# so the same build artifact works correctly on both HTTP staging and HTTPS production.
+# Set to `true` on any deployment that is served over HTTPS.
+# If unset, defaults to `true` when NODE_ENV=production and `false` otherwise.
+COOKIE_SECURE=false
```
Oiranca: Setting this to `false` in `.env.example` is risky because teams often copy example env files into staging/production. Since the runtime default is already secure in `NODE_ENV=production` when the variable is unset, the safer example is to leave `COOKIE_SECURE` commented out (or document `COOKIE_SECURE=false` only in a local-dev override). Otherwise a deployed HTTPS environment can accidentally ship auth and CSRF cookies without the `Secure` flag.

Oiranca: ✅ Fixed. `COOKIE_SECURE=false` is no longer active in `.env.example` — it is now commented out with documentation. The block explains the flag is evaluated at runtime, defaults to `true` when `NODE_ENV=production` (and `false` otherwise), and should be left unset in production so an HTTPS deployment never accidentally ships auth/CSRF cookies without the `Secure` flag. Only a commented `# COOKIE_SECURE=false` remains, for local HTTP dev opt-in.

lib/server/security.ts
```diff
@@ -373,10 +353,100 @@ export function enforceSameOriginForMutation(request: NextRequest): NextResponse
...
+  }
+
+  return ratelimit
+}
+
+async function enforceRateLimitRedis(
+  request: NextRequest,
+  policy: RateLimitPolicy,
+): Promise<NextResponse | null> {
+  const ratelimit = await getRatelimitForPolicy(policy)
```
Oiranca, line 426: The Redis-backed rate-limit path is the production security path, but the tests only exercise the in-memory fallback. Please mock `@upstash/redis` and `@upstash/ratelimit` with `UPSTASH_REDIS_REST_URL/TOKEN` set and assert both allowed and blocked requests, including the `Retry-After` behavior. Otherwise a broken dynamic import, bad duration format, or changed Upstash result shape can ship while coverage stays green.

Oiranca (commented):
Review pass: security hardening looks directionally right; missing Redis-path test coverage should be added before relying on the shared limiter in production.

lib/server/security.ts
Oiranca: Added Redis-backed rate-limit coverage in commit 49ab7dc (`__tests__/server/security.test.ts`).

The new `describe('Redis-backed rate limiting (Upstash)')` block sets `UPSTASH_REDIS_REST_URL`/`UPSTASH_REDIS_REST_TOKEN` and mocks `@upstash/redis` and `@upstash/ratelimit` (the mock is registered via `vi.mock`, which is hoisted above imports, so it intercepts the dynamic `import()` calls in `getRatelimitForPolicy`). Three cases are covered:

1. **Allowed request** — `ratelimit.limit()` resolves `{ success: true, ... }`, `enforceRateLimit` returns `null`, and we assert the `Redis` constructor + `Ratelimit.slidingWindow(policy.limit, \`${policy.windowMs}ms\`)` were called with the expected arguments (this pins the sliding-window duration string format).
2. **Blocked request** — `ratelimit.limit()` resolves `{ success: false, reset: <future ms> }`, and we assert the response is a 429 with a `Retry-After` header correctly derived from Upstash's `reset` timestamp.
3. **Singleton reuse** — two requests against the same policy only construct the `Redis` client and `Ratelimit` instance once (module-level cache), while `limit()` is called per-request.

I verified the new tests actually catch regressions by temporarily changing the sliding-window duration format from `` `${policy.windowMs}ms` `` to `` `${policy.windowMs}` `` (dropping the unit suffix) — the "allowed" test failed with a clear assertion diff, then I reverted the change (confirmed via `git diff` showing no diff on `lib/server/security.ts`) and re-ran the full suite green (15/15).

PR #142, opened 2026-06-24
fix(cron): register mark-no-show on Vercel Cron + GET handler [KIM-404]

Oiranca:
## Summary

`vercel.json` only registered the deprecated `cancel-pending` route (returns 410). The active `app/api/cron/mark-no-show` route was unregistered, unpinned, and had no `maxDuration`.

### Changes
- **`vercel.json`**: drop the dead `cancel-pending` entry; register `/api/cron/mark-no-show` as a Vercel Cron (`*/15 * * * *`); pin `regions: ["arn1"]` (Stockholm, matching the Supabase eu-north-1 region); set `maxDuration: 60` for the cron function; add `framework: "nextjs"`.
- **`app/api/cron/mark-no-show/route.ts`**: add a `GET` handler. **Vercel Cron invokes the registered path via GET**, but the route only exported `POST` — it would have returned 405 and the cron would never run. The new `GET` delegates to the same `handleCronRequest`, so the `Authorization: Bearer $CRON_SECRET` check (which Vercel Cron satisfies automatically) still gates it. `POST` is kept unchanged.

## Validation

- `pnpm typecheck` — pass
- `pnpm test` — **675/675** pass (mark-no-show: 6 POST + 6 GET), 0 skipped
- `pnpm build` — pass

## Follow-up for the user
Set `CRON_SECRET` in Vercel project env so Vercel Cron's requests authenticate. Once this deploys, the external cron-job.org trigger can be retired.

Closes KIM-404

### DIFF
--- a/__tests__/app/api/cron/mark-no-show.test.ts
+++ b/__tests__/app/api/cron/mark-no-show.test.ts
@@ -111,4 +111,88 @@ describe('/api/cron/mark-no-show', () => {
       expect(await response.json()).toEqual({ marked: 0 })
     })
   })
+
+  describe('GET method', () => {
+    it('returns 401 when Authorization header is missing', async () => {
+      vi.stubEnv('CRON_SECRET', 'test-secret')
+      const { GET } = await import('@/app/api/cron/mark-no-show/route')
+
+      const request = createRequest('/api/cron/mark-no-show', 'GET')
+      const response = await GET(request)
+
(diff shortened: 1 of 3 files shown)


Oiranca:
Closing KIM-404. Vercel deployment check failing and release already shipped via PR #155. Can reopen if cron work is picked up again.

PR #143, opened 2026-06-24
test(coverage): expand coverage to service + API layer [KIM-406]

Oiranca:
## Summary

`vitest.config.mts` `coverage.include` only watched 6 auth-layer files, so the 85% thresholds never applied to reservations / equipment / events / rooms / saved-games. This widens the lens to the whole service + API layer and sets honest, regression-catching thresholds.

### Changes (`vitest.config.mts` only)

- **`coverage.include`** → `lib/server/**/*.ts`, `app/api/**/*.ts`, `lib/auth/auth-context.tsx`; exclude `**/*.d.ts`.
- **Global thresholds** set to the measured floor: lines 80, statements 80, functions 85, branches 75 (actual: 80.57 / 80.57 / 88.36 / 76.24). This makes genuinely untested endpoints a *visible* gap instead of hiding them behind a narrow include.
- **Per-glob overrides** keep the security-critical auth layer at its real, high level so it can't silently slide to the global floor:
  - `lib/server/auth.ts` → 100/100/93/100
  - `lib/server/auth-service.ts` → 85/85/63/85
  - `app/api/auth/**/*.ts` → 87/85/75/87

The 0%-coverage endpoints (equipment, reservations, rooms, saved-games, users CRUD) are now reported and tracked for follow-up coverage work; this PR is config only and adds no tests.

## Validation

- `pnpm test:coverage` — **green** (global + per-glob thresholds all met)
- `pnpm test` — 669/669 pass, 0 skipped
- `pnpm typecheck` — pass
- `pnpm build` — pass

Closes KIM-406

### DIFF
--- a/vitest.config.mts
+++ b/vitest.config.mts
@@ -25,20 +25,41 @@ export default defineConfig({
     coverage: {
       provider: 'v8',
       reporter: ['text', 'lcov'],
-      exclude: ['node_modules', '.next'],
+      exclude: ['node_modules', '.next', '**/*.d.ts'],
       include: [
-        'app/api/auth/**/*.ts',
+        'lib/server/**/*.ts',
+        'app/api/**/*.ts',
         'lib/auth/auth-context.tsx',
-        'lib/server/auth.ts',
-        'lib/server/auth-service.ts',
-        'lib/server/http-error.ts',
-        'lib/server/service-error.ts',
       ],
+      // Global thresholds: baseline floor across all measured files (80.57% actual lines).
+      // Untested API routes (0% coverage) are allowed at global floor; they must be
+      // improved separately as part of KIM-408+. Per-glob overrides below preserve
+      // the high bar for security-critical auth layer so regressions are caught.
       thresholds: {
-        lines: 85,
+        lines: 80,
         functions: 85,
         branches: 75,
-        statements: 85,
+        statements: 80,
+        // Auth layer: measured at 85-100% lines across all routes.
+        // Per-glob overrides hold the strong bar here so regressions trigger alerts.
+        'lib/server/auth-service.ts': {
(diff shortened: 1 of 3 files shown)


vitest.config.mts
```diff
@@ -25,20 +25,41 @@ export default defineConfig({
     coverage: {
       provider: 'v8',
       reporter: ['text', 'lcov'],
-      exclude: ['node_modules', '.next'],
+      exclude: ['node_modules', '.next', '**/*.d.ts'],
       include: [
-        'app/api/auth/**/*.ts',
+        'lib/server/**/*.ts',
+        'app/api/**/*.ts',
```
Oiranca, line 31: This expands the `include` list, but Vitest does not report unimported source files as 0% covered unless coverage is configured to include all matching files. That means many `app/api/**/*.ts` routes can still be absent from the coverage denominator, so the new thresholds may give a false sense of API coverage. Please enable the equivalent of `coverage.all: true` for these globs (or add an explicit import smoke test per route) and include a coverage output sample proving untested routes are counted.

Oiranca: ✅ Addressed. Set `coverage.all: true` explicitly in `vitest.config.mts` so every file matched by the `include` globs (incl. untested `app/api/**/*.ts` routes) is counted in the denominator instead of being silently dropped. Coverage evidence: all 32 `app/api/**/route.ts` files and all 17 `lib/server/**` files now appear in the report — untested routes show as 0% rather than being absent, e.g. `app/api/rooms/route.ts 0% (1-34)`, `app/api/reservations/route.ts 0% (1-45)`. (Note: Vitest 3.2.4 already defaults `all` to true; making it explicit guards against an implicit default silently changing and documents intent.)

PR #145, opened 2026-06-24
docs(security): justify admin-client usage in member services [KIM-405]

Oiranca:
## Summary

KIM-405 originally proposed switching member-scoped reads/inserts to the RLS-scoped session client. After review (confirmed with the maintainer), that switch is **intentionally not made**:

- The migration's locked decision is to **eliminate RLS** (Phase 2). The session client depends on RLS, so a switch would be reverted shortly and contradicts the current model.
- **KIM-397** already added `assertMemberRowsScoped()` as the app-layer defense for member reads, specifically because RLS is going away.

So the acceptance criterion ("admin client only where justified; comment explaining each use") is satisfied by **auditing and documenting** each admin-client usage — confirming it is deliberate — with **zero behavior change**.

## Changes (comments only)

- `lib/server/saved-games-service.ts`: module-level policy header + a justification comment at each of the 6 admin-client usages (member reads → `user_id` filter + `assertMemberRowsScoped`; inserts → explicit `user_id: session.id`; renew → ownership check throwing 403; attendance recording → system/cron cross-user context).
- `lib/server/reservations-service.ts`: justification on the `listVisibleReservations` member-read path (`effectiveUserId` derived from `session.id` for non-admins).

## Validation

- `pnpm typecheck` / `pnpm lint` / `pnpm build` — pass
- `pnpm test` — **669/669** pass
- Diff confirmed comments-only (no logic/signature changes)

Closes KIM-405

### DIFF
--- a/lib/server/saved-games-service.ts
+++ b/lib/server/saved-games-service.ts
@@ -6,9 +6,11 @@
  * the session-scoped client cannot be relied upon for row-level isolation.
  * Member isolation is instead enforced at the application layer via:
  *   - explicit `user_id = session.id` filters on member-scoped reads/writes
- *   - assertMemberRowsScoped() (KIM-397) called on any multi-row member read
+ *     (see the non-admin branch of `listSavedGamesForSession` below)
  *   - explicit ownership checks (`current.user_id !== session.id`) before
  *     mutations such as renew
+ * NOTE: a stronger multi-row guard (`assertMemberRowsScoped()`, KIM-397) is
+ * planned but not part of this PR — it is not imported or called here today.
  * Cross-user operations (attendance recording, table/event conflict checks)
  * legitimately need to span users and therefore require the admin client.
  */
--- a/lib/server/saved-games-service.ts
+++ b/lib/server/saved-games-service.ts
@@ -1,3 +1,17 @@
+/**
+ * Saved-games service — admin client usage policy
+ *
+ * This service uses createSupabaseServerAdminClient() throughout deliberately.
+ * RLS is being removed as part of the Vercel/Postgres migration (Phase 2), so
+ * the session-scoped client cannot be relied upon for row-level isolation.
+ * Member isolation is instead enforced at the application layer via:
+ *   - explicit `user_id = session.id` filters on member-scoped reads/writes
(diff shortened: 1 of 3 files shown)


lib/server/saved-games-service.ts
```diff
@@ -1,3 +1,17 @@
+/**
+ * Saved-games service — admin client usage policy
+ *
+ * This service uses createSupabaseServerAdminClient() throughout deliberately.
+ * RLS is being removed as part of the Vercel/Postgres migration (Phase 2), so
+ * the session-scoped client cannot be relied upon for row-level isolation.
+ * Member isolation is instead enforced at the application layer via:
+ *   - explicit `user_id = session.id` filters on member-scoped reads/writes
+ *   - assertMemberRowsScoped() (KIM-397) called on any multi-row member read
```
Oiranca: This says `assertMemberRowsScoped()` protects multi-row member reads, but this PR does not include that import/call, and `listSavedGamesForSession` below still maps `data` directly. If #145 can merge independently into `develop`, this comment documents a security guarantee that does not exist. Please either stack/rebase this on the KIM-397 guard change so the code and comment land together, or scope the documentation to the protections actually present in this PR.

Oiranca (commented):
Review pass: documentation-only PR currently overstates an isolation guarantee that is not present in this branch.

lib/server/saved-games-service.ts
Oiranca: Good catch — thanks for flagging this. Since this PR is docs-only and shouldn't take on a dependency on KIM-397 (PR #138) landing first, I corrected the doc comment instead of stacking on it.

The comment no longer claims `assertMemberRowsScoped()` is in effect (it isn't imported or called anywhere in this file today). It now only describes the protection actually present in this PR: the `user_id = session.id` filter in the non-admin branch of `listSavedGamesForSession`, plus explicit ownership checks before mutations like renew. A NOTE was added clarifying the stronger multi-row guard is planned under KIM-397 but not part of this PR.

Fixed in commit 871774c (comment correction) — see also 22c3bdf for an unrelated local pre-push lint fix pushed in the same batch. This PR should now be safe to merge independently into `develop` without implying a guarantee that doesn't exist yet.

PR #146, opened 2026-06-24
refactor(users): extract member-import parser into its own module [KIM-407]

Oiranca:
## Summary

`lib/server/users-service.ts` (~872 lines) was a god-file mixing user CRUD with ~420 lines of member-import parsing (CSV/ODT/XLSX). This extracts all parsing/normalization into a new **`lib/server/member-import.ts`**; `users-service.ts` keeps only the DB orchestration.

- **Moved** to `member-import.ts`: `parseCsv`, `detectDelimiter`, `normalizeHeader`, `extractSpreadsheetCsv` (exceljs), `extractOdtCsv`, `rowsToCsv`, `buildCanonicalMemberImportCsv`, `assertSourceArchiveMatchesExtension`, `parseMemberImportCsv`, `normalizeMemberImportSource`, etc. — plus their types/constants and the `exceljs`/`fflate` imports.
- **Stayed** in `users-service.ts`: all user CRUD and the DB import orchestration (`importMembersFromCsv`, `importMembersFromSource`, `importMembersFromNormalizedRows`), which now import the parsing entry points from the new module.
- **Pure move** — no behavior changes. `users-service.ts`: **872 → 449 lines**.
- The member-import test was repointed to import the parsing functions from `@/lib/server/member-import`.

## ⚠️ Stacked on #136 (KIM-395)

This branch is based on the **KIM-395** branch so the extracted module uses `exceljs` (consistent with the security fix), not the removed `xlsx`. **Merge #136 first**, then this. Once #136 is in `develop`, this PR's diff narrows to just the extraction.

## Validation

- `pnpm typecheck` / `pnpm lint` — pass
- `pnpm test` — **669/669** pass (21 member-import tests: CSV/ODT/XLSX + orchestration), 0 skipped
- `pnpm build` — pass

Closes KIM-407


lib/server/member-import.ts
```diff
@@ -0,0 +1,435 @@
...
+  const wb = new ExcelJS.Workbook()
+
+  try {
+    await wb.xlsx.load(bytes.buffer as ArrayBuffer)
+  } catch {
+    serviceError('Spreadsheet file is invalid or corrupted', 400)
+  }
+
+  if (wb.worksheets.length === 0) {
+    serviceError('Spreadsheet does not contain any sheets', 400)
```
Oiranca, line 195: Same sliced-buffer issue here: `bytes.buffer` is the full backing `ArrayBuffer`, not necessarily the uploaded byte range. If `normalizeMemberImportSource` receives a `Uint8Array` view with non-zero `byteOffset`, ExcelJS sees the padding bytes too and can reject a valid XLSX. Use `bytes.buffer.slice(bytes.byteOffset, bytes.byteOffset + bytes.byteLength)` and cover that case in the parser tests.

Oiranca: ✅ Fixed in `4c276d5`. `normalizeMemberImportSource` / `extractSpreadsheetCsv` now passes `bytes.buffer.slice(bytes.byteOffset, bytes.byteOffset + bytes.byteLength)` to ExcelJS instead of the full backing `ArrayBuffer`. Added a regression test that wraps a valid XLSX in a `Uint8Array` view with non-zero `byteOffset` plus a crafted trailing fake zip EOCD record — verified it throws `Spreadsheet does not contain any sheets` on the old code and passes with the fix.

PR #148, opened 2026-07-03
OIR-202: public marketing landing page for unauthenticated visitors

Oiranca:
## Summary

Two pieces of work, both validated by qa-engineer and security-reviewer:

1. **Architecture fold-in** — consolidated the originally-proposed standalone `club_events` table into the existing `public.events` table instead, extending it with bilingual columns (`title_es/en`, `blurb_es/en`, `description_es/en`, `date_kind`, `end_date`, `recurrence_label_es/en`, `image_url`, `link_url`, `category_es/en`). Adds RLS policy `events_select_public` (anon `SELECT` gated on `title_es`/`title_en` both `IS NOT NULL`) plus a scoped `GRANT SELECT ON TABLE public.events TO anon` (required because an earlier migration had revoked anon's base table grant, which would otherwise make the new policy dead code). Seeds 24 real Alea Las Palmas club events. `lib/server/club-events-service.ts` reads from `events`, filters to bilingual rows only, and derives upcoming/past status from date at read time (no stored flag).
2. **Visual-fidelity rebuild** — rebuilt `components/landing/*` and the landing render in `app/[locale]/page.tsx` to match the real design source: dark theme with custom tokens, custom fonts via `next/font`-style imports, sticky nav, interactive D20 dice-roll widget, marquee-style scrolling carousels (events/games/partners), event detail modal, footer, and a seeded-PRNG hex-grid background (SSR-hydration safe). Real image assets wired in from `alealaspalmas.es`.

## Update 2026-07-04 — design parity + DB repair

- **`fix(db)`: repair migration `20260704000001_oir202_repair_events_anon_grant.sql`** — the remote DB had recorded `20260703000001` as applied while its anon `GRANT`/policy portion never took effect (migration-history rewrite on this branch: columns and seed were present, but anon `SELECT` failed with `42501`, hard-500ing the public landing). This standalone idempotent migration re-applies only `GRANT SELECT ON public.events TO anon` and recreates `events_select_public`. Root cause confirmed by a direct anon-key REST probe.
- **`fix(landing)`: design-parity die cursor** — `custom-cursor.tsx` rendered a hardcoded pawn icon in `#c8a25b`; the design source mounts `<CustomCursor variant="die" color="#e6c281" />`. Component now supports `die`/`meeple`/`pawn` variants and the landing mounts the gold die per the design. Trailing ring, lerp factor, scoping and touch fallback were verified already matching.
- **`feat(landing)`: full easter-egg suite + Tailwind design tokens** — ported from the designer's official `alea-nextjs` reference port (same stack: Next.js + TS + Tailwind + shadcn/ui): nat-20 modal (`easter-egg.tsx`), canvas dice rain (`dice-rain.tsx`), catchable meeple with `MEEPLE26` reward modal (`meeple-hunt.tsx`/`meeple-egg.tsx`), `use-shake`/`use-tap-count` hooks, hero D20 `alea:nat20` dispatch, footer hint. i18n keys `home.egg.*`/`home.meeple.*`/`home.easter.hint` added to both locales with full parity. `tailwind.config.ts` gains the `alea` palette and `mod-pulse`/`mod-marq`/`mod-float` keyframes as additive tokens for stack consistency.

## Security review (this PR's final gate)

- **RLS**: `events_select_public` correctly excludes internal admin-created room-booking events (which only ever populate the legacy single-locale `title` column, never `title_es`/`title_en`). No write path exists today for anon/authenticated users to set `title_es`/`title_en` directly — writes only go through `SECURITY DEFINER` RPCs granted to `service_role` and admin-checked in the service layer — so there is no reachable bypass of the public/private boundary pre-admin-CRUD. The `GRANT SELECT` is scoped to `SELECT` only and does not reinstate broader anon privileges revoked by an earlier migration (KIM-383). `event_room_blocks` is untouched — still fully revoked from `anon`.
- **Migrations**: additive-only, safe for a single fresh apply; committed but intentionally **not executed** by any agent (user applies via `supabase db push`, per project rules). Seed data verified to contain only real public event-marketing copy, no secrets, no unexpected PII, and SQL string literals verified quote-balanced (no injection risk).
- **XSS**: no `dangerouslySetInnerHTML` anywhere in the new code; all DB-sourced text renders via plain JSX (React default-escaped).
- **Secrets**: none found in the diff.

## Follow-ups (deferred, non-blocking)

- **Admin CRUD UI** for managing these public club events — explicitly out of scope for this issue (see `docs/issues/oir-202-public-landing-page.md`).
- **MEDIUM**: `image_url`/`link_url` are rendered as raw `src`/`href` with no protocol allowlist. Not exploitable today (no untrusted write path exists), but should be hardened (restrict to `http(s):`) before the admin-CRUD UI ships and admins can freely edit these fields.
- **LOW**: `app/[locale]/page.tsx` calls `listClubEvents()` with no `try/catch` — a DB error currently hard-500s the whole public landing route instead of degrading gracefully. Recommend a quick follow-up.
- ~~Full Konami/shake/tap easter-egg suite from the design source was intentionally deferred~~ — **landed in this PR** (2026-07-04 update above). Only `useKonami` remains unported, matching the design's own App shell which never wired it.
- The admin dashboard's existing `listEvents()` will now also surface these 24 landing rows (no room blocks) — flagged in `docs/issues/oir-202-public-landing-page.md` for the admin-CRUD follow-up to address.

## Test plan

- [x] `pnpm typecheck` / `pnpm lint` / `pnpm build` all green (confirmed by qa-engineer and again on push via pre-push hook, re-run after each 2026-07-04 commit)
- [x] `pnpm test`: 670/670 passing (51/51 files), including updated auth-page redirect expectations
- [x] i18n key parity between `messages/en.json` and `messages/es.json` (extended with egg/meeple/easter keys, parity maintained)
- [x] Anon REST probe against remote confirmed root cause of the landing 500 (`42501 permission denied for table events`)
- [ ] User applies `supabase db push` for the **three** migrations (extend + seed + `20260704000001` repair), then manually verifies `/es` and `/en` render the landing for unauthenticated visitors (including custom die cursor and easter eggs) and that authenticated visitors still redirect to `/rooms`


lib/hooks/use-shake.ts
```diff
@@ -0,0 +1,62 @@
...
+          count = 0
+        }
+      }
+      last = { x: acc.x || 0, y: acc.y || 0, z: acc.z || 0, t: now }
+    }
+
+    const attach = () => window.addEventListener('devicemotion', handler)
+
+    // iOS requires explicit permission; request it on the user's first touch.
+    const DME = DeviceMotionEvent as unknown as {
```
Oiranca, line 45: This reads the global `DeviceMotionEvent` before the `typeof DeviceMotionEvent !== 'undefined'` guard below. In browsers/environments where that API is missing, the effect throws a `ReferenceError` as soon as the landing mounts, instead of silently disabling shake detection. Please move this cast inside the guarded branch, e.g. check `typeof DeviceMotionEvent === 'undefined'` first and return/skip attaching before referencing the constructor.

Oiranca (commented):
Review pass: one client-side robustness issue found in the landing easter-egg hook.

lib/hooks/use-shake.ts
Oiranca: Fixed in 98b4926 ✅ — reordered the guard so the `typeof DeviceMotionEvent === 'undefined'` check runs first and returns early before any reference to the `DeviceMotionEvent` global. The cast to `{ requestPermission?: ... }` (and the subsequent `typeof DME.requestPermission === 'function'` check) now only executes once we've confirmed the API exists, so environments without `DeviceMotionEvent` no longer throw a `ReferenceError` on mount — shake detection is silently skipped instead.

PR #149, opened 2026-07-04
OIR-203: admin CRUD for public club events (stacked on OIR-202)

Oiranca:
## Summary

- Admin CRUD (create/read/update/delete) for public "club events" (the tournament/game-night/club-history content shown on the OIR-202 landing page), backed by `lib/server/club-events-service.ts` and new routes `app/api/club-events` / `app/api/club-events/[id]`.
- Room blocking is optional and, when attached, is applied atomically via a new SECURITY DEFINER RPC `apply_club_event_room_blocks` (migration `20260704000004_oir203_apply_club_event_blocks_rpc.sql`) — replaces the room-block set for an event, cancels overlapping active/pending reservations, all in one transaction (mirrors the existing `update_event_with_blocks` pattern).
- Shared absolute http(s) URL allowlist extracted to `lib/validations/url.ts` and applied to `image_url`/`link_url` on both create and update — `javascript:`/`data:`/relative URLs are rejected before persistence.
- New "Club events" section in the admin dashboard (`components/admin/club-events-section.tsx`) with Upcoming/Past tabs, wired through `lib/hooks/use-admin.ts`.
- Legacy-surface guard (`isClubEventRow`): the internal room-booking `updateEvent`/`deleteEvent` flow now rejects (404) any row that has become public landing content (both `title_es`/`title_en` populated), and the club-events service enforces the inverse — each surface only ever operates on rows it owns.
- New `events_bilingual_titles_paired` CHECK constraint: `title_es`/`title_en` must both be null or both be populated, so a row can never be "half" landing content. Verified safe against the existing seeded club-event rows and all internal room-booking rows.

## Stacked PR

This branch is stacked on `feat/oir-202-public-landing-page` (PR #148). The diff shown against `develop` includes the OIR-202 changes as well as OIR-203. Merge the **last** PR in the chain into `develop` to bring in the whole chain in one go — do not merge this PR standalone before #148.

## Security review (security-reviewer, final gate)

- `apply_club_event_room_blocks`: `SECURITY DEFINER`, `search_path` pinned to `public, pg_catalog`, `EXECUTE` revoked from `PUBLIC`/`anon`/`authenticated`, granted only to `service_role`. All jsonb payload fields are read via `->>` and cast to typed columns (uuid/date/time/boolean) inside parameterized `INSERT`/`UPDATE` statements — no dynamic SQL, no injection surface.
- All `club-events` routes run `enforceMutationSecurity` + `enforceRateLimit` (admin-mutation policy) + `requireAdmin` before touching the request body; all authorization (`role === 'admin'`) and ownership logic lives in the service layer (`requireAdminSession`), never in the route handlers.
- URL allowlist (`lib/validations/url.ts`, http(s)-only) is applied to `imageUrl`/`linkUrl` on both `createClubEvent` and `updateClubEvent` via the shared `resolveClubEventFields` path.
- `isClubEventRow` guard confirmed bidirectional: legacy `updateEvent`/`deleteEvent` reject club-event rows (404), and `updateClubEvent`/`deleteClubEvent` reject non-club-event rows (404) — no cross-surface leakage.
- No secrets, credentials, or PII found in the diff. i18n keys (`messages/en.json`/`es.json`) verified at full parity. No `dangerouslySetInnerHTML` or `eval` introduced.
- Verdict: **APPROVE** — no blocking findings.

## Test plan

- 698/698 tests passing (qa-engineer re-approval after code-review fixes, round 2).
- `typecheck`, `lint`, and `build` all green.
- **Migration required**: this PR includes a new migration file. The user must run `supabase db push` manually to apply it — no agent has applied or will apply it.


lib/server/club-events-service.ts
```diff
@@ -0,0 +1,581 @@
...
+
+  // Finding 2: validate EVERYTHING — fields, URL allowlist, and (when
+  // blocksRooms is set) the schedules payload — before any DB write.
+  const fields = resolveClubEventFields(body, null)
+  const wantsBlocks = parseBooleanFlag(body.blocksRooms)
+  const schedules = wantsBlocks ? validateSchedulesPayload(body.schedules) : null
+
+  const admin = createSupabaseServerAdminClient()
+
+  const { data, error } = await admin
```
Oiranca, line 501: This creates the `events` row before the block-replacement RPC. If `apply_club_event_room_blocks` fails after the insert (bad room id, FK issue, DB error, transient failure), the API returns an error but leaves a new club event persisted with no intended room blocks/cancellations. Please make the event insert and block creation atomic, either by moving create+blocks into one SQL/RPC transaction or by validating all referenced rooms before the insert and adding a regression test that forces the block RPC to fail and proves no event row is left behind.

Oiranca: ✅ Fixed in `37fdb05`. `createClubEvent` can no longer leave an orphan event row: (1) all referenced `room_id`s are validated against the `rooms` table **before** the `events` insert (unknown room → 400 up front), and (2) the `apply_club_event_room_blocks` call is wrapped in try/catch with a **compensating delete** of the just-inserted event row on any RPC failure, rethrowing the original error. Added 2 regression tests: one forces the block RPC to fail and asserts the event row is deleted; one forces the room-existence check to fail and asserts no `events` insert happens at all. Full suite green (700 tests).

Oiranca: Follow-up (post QA + security review, `0d407c1`): hardened the compensating-delete path flagged in review. The delete result is now error-checked and, on failure, logs `[club-events] compensating delete failed … orphaned event row requires manual cleanup: <id>` (matching the uploads-service logging style) while still rethrowing the ORIGINAL RPC error — so a failed cleanup can no longer silently leave a publicly-visible orphan event without ops visibility. Added a regression test forcing BOTH the block RPC and the compensating delete to fail (asserts original error rethrown + console.error with the orphan id). Note: a full single-transaction `create_club_event_with_blocks` RPC remains the gold-standard option, but the room-pre-validation + logged-compensation fallback (explicitly accepted in the original comment) is what's shipped here.

Oiranca (commented):
Review pass: create-path orphan handling is fixed; update path still has a partial-write failure mode.

lib/server/club-events-service.ts
```diff
@@ -0,0 +1,636 @@
...
+
+  const blocksRoomsProvided = body.blocksRooms !== undefined
+  const schedulesProvided = body.schedules !== undefined
+  const wantsBlocks = blocksRoomsProvided ? parseBooleanFlag(body.blocksRooms) : undefined
+
+  const validatedSchedules = wantsBlocks !== false && schedulesProvided
+    ? validateSchedulesPayload(body.schedules)
+    : null
+
+  const { data, error } = await admin
```
Oiranca, line 589: `updateClubEvent` writes the event fields before applying the new room blocks. If the later block RPC fails (bad room id, FK issue, transient DB error), the request returns an error but the metadata update has already been committed, leaving the event in a partially updated state. Please validate referenced rooms before the UPDATE and/or move field update + block replacement into one transaction/RPC; add a regression test that forces block replacement to fail and asserts the original event fields remain unchanged.

Oiranca: Fixed in a2a59a6. `updateClubEvent` now validates every room id referenced in `schedules` against the `rooms` table (reusing the same `validateRoomsExist` helper already used by `createClubEvent` for the analogous PR #149 finding) BEFORE the event-fields UPDATE runs, so a bad room id is rejected with 400 and zero writes happen.

As a second layer, if the block-replacement RPC still fails for some other reason (transient DB error, etc.) after room ids passed validation, the fields UPDATE is now reverted back to its pre-update values (a compensating UPDATE), so the event never ends up in a partially-updated state — mirroring the compensating-delete pattern used in `createClubEvent`. If that compensating revert itself fails, it's logged loudly (without masking the original error returned to the client) so the row can be reconciled manually.

Added regression tests in `__tests__/server/club-events-service.test.ts`:
- rejects an unknown room id with 400 before any UPDATE runs
- forces the block RPC to fail and asserts the fields UPDATE is reverted to the original values
- forces both the RPC and the compensating revert to fail, asserting the original RPC error is still what's rethrown, with the failure logged

PR #153, opened 2026-07-04
OIR-207: device image uploads to Supabase Storage (stacked on OIR-206)

Oiranca:
## Summary

- Adds device image uploads to Supabase Storage for landing content (club events, partners, library games) via a new public `landing-media` bucket and an admin-only upload route.
- New migration `20260704000005_oir207_landing_media_bucket.sql`: creates the `landing-media` bucket (public read, 5 MB `file_size_limit`, 4-MIME `allowed_mime_types` allowlist at the bucket level), and a single SELECT-only storage policy scoped to `bucket_id = 'landing-media'` for `anon`/`authenticated` (idempotent via `DROP POLICY IF EXISTS`). No client write policies exist — all writes go through the `service_role` client. `library_games.img_url` is added as a nullable column guarded with `ADD COLUMN IF NOT EXISTS`.
- New `lib/server/uploads-service.ts`: admin-only (`role === 'admin'`) upload service. Folder is restricted to a strict allowlist (`events` | `partners` | `library-games`); MIME type is restricted to `image/png`, `image/jpeg`, `image/webp`, `image/gif` (SVG intentionally excluded — XSS/script-execution vector); the stored file extension is derived from the validated MIME type map, never from the caller-supplied filename; the stored path is `${folder}/${randomUUID()}.${ext}` with no user-controlled path segments. Storage errors are logged server-side only; the client always receives a generic message.
- New `app/api/admin/uploads/route.ts`: `enforceMutationSecurity` + rate limiting (`adminMutation` policy) + `requireAdmin` all run before the multipart body is parsed; a `formData()` parse failure returns 400.
- New shared `components/admin/image-upload.tsx` control wired into the club-events, partners, and library-games admin forms, alongside the existing manual URL input (kept as an editable fallback).
- `library_games` gains an optional `imgUrl`/`img_url` field end-to-end (service, types, admin form); `GameCard` renders it with a gradient-cover fallback plus a scrim overlay (`.mod-game-img-scrim`) for text legibility/contrast when an image is present.
- `imageUrl` on library games goes through the existing shared `validateOptionalUrl` http(s)-only allowlist (same validator already used by events/partners).

## Stacked chain note

**This is the NEW FINAL PR of the stacked chain #148 → #149 → #150 → #151 → #152 → this PR.**

The user will merge **only** `feat/oir-207-image-uploads` into `develop` (`git merge --no-ff` + push) — all prior chain PRs (#148–#152) will auto-close as their commits are already included in this branch's history.

## Security summary

- Bucket is public-read only; no client-side write/insert/update/delete storage policy exists — writes exclusively go through the `service_role` client, bypassing RLS.
- Upload route: `requireAdmin` (admin-role gate) runs before any request body is parsed, in addition to CSRF/mutation-security enforcement and per-route rate limiting.
- MIME allowlist excludes SVG (confirmed — SVG can execute embedded scripts/JS in some contexts).
- Stored object extension is derived only from the validated MIME map, never from the client-supplied filename.
- Stored object path has no user-controlled segments (`${folder}/${randomUUID()}.${ext}`), folder itself is allowlist-validated server-side.
- Storage errors are logged server-side; only a generic 500 is ever returned to the client.
- Upload requests go through the shared `apiClient` wrapper, so CSRF headers are attached the same as any other admin mutation (verified `FormData` bodies still get the CSRF header and skip the JSON `Content-Type` override).
- Rendered images use plain `` tags (no `dangerouslySetInnerHTML`); the game-card image sits behind a scrim overlay; `imageUrl`/`img_url` fields for events/partners/library-games all continue to pass through the shared absolute-http(s)-only URL validator before persistence.
- i18n parity confirmed for all new `imageUpload.*` keys in `messages/en.json` and `messages/es.json`.
- No secrets, credentials, or API keys introduced in this diff.

## Test plan

- [x] 836/836 tests passing (`npx vitest run`), including new coverage in `__tests__/server/uploads-service.test.ts` and updates to `__tests__/server/library-games-service.test.ts`.
- [x] `tsc --noEmit` clean.
- [ ] After merge to `develop`, run `supabase db push` to apply migrations `20260703*` through `20260704000001`–`20260704000005` (includes this PR's `20260704000005_oir207_landing_media_bucket.sql`).


lib/server/uploads-service.ts
```diff
@@ -0,0 +1,113 @@
...
+  if (typeof folder !== 'string' || !ALLOWED_FOLDERS.has(folder as UploadFolder)) {
+    serviceError('folder must be one of: events, partners, library-games', 400)
+  }
+  return folder as UploadFolder
+}
+
+function requireValidFile(file: UploadFileLike | null): { file: UploadFileLike; extension: string } {
+  if (!file) serviceError('file is required', 400)
+
+  const extension = MIME_TO_EXTENSION[file.type]
```
Oiranca, line 125: This trusts the client-provided MIME type as the only file-type check. Because the object is written to a public bucket and later rendered on the landing page/admin previews, the service should also verify the decoded magic bytes or use an image decoder/probe before upload. Add negative tests where the `File.type` is `image/png`/`image/jpeg` but the body is not that format, and reject those before writing to Storage.

Oiranca: ✅ Fixed in `4f3587f`. The upload service no longer trusts the client `File.type` alone. Before writing to Storage, `uploadLandingMediaImage` now verifies the decoded **magic bytes** of the body and requires them to match both an allowed image type AND the declared MIME (rejects cross-format spoofing). Signatures checked: PNG `89 50 4E 47 0D 0A 1A 0A`, JPEG `FF D8 FF`, WEBP `RIFF…WEBP`, GIF `GIF87a/GIF89a`. Rejections reuse the existing 400 `serviceError` (no new error shape), and nothing is written to Storage on failure. Added negative tests (declared `image/png`/`image/jpeg` with non-matching bodies, and cross-format png-declared/jpeg-body) asserting rejection + no storage write, plus positive matching-signature tests. 31/31 passing.

supabase/migrations/20260704000005_oir207_landing_media_bucket.sql
```diff
@@ -0,0 +1,33 @@
...
+-- ON CONFLICT DO NOTHING keeps this migration idempotent/re-runnable.
+INSERT INTO "storage"."buckets" ("id", "name", "public", "file_size_limit", "allowed_mime_types")
+VALUES (
+  'landing-media',
+  'landing-media',
+  true,
+  5242880, -- 5 MB, mirrors the service-layer limit enforced in uploads-service.ts
+  ARRAY['image/png', 'image/jpeg', 'image/webp', 'image/gif']
+)
+ON CONFLICT ("id") DO NOTHING;
```
Oiranca: Because this is security-relevant bucket configuration, `ON CONFLICT DO NOTHING` can leave a pre-existing `landing-media` bucket with stale settings (for example no 5 MB limit, a broader MIME allowlist, or `public = false`) while the migration still succeeds. Please make the migration converge the bucket to the intended state with `ON CONFLICT (id) DO UPDATE SET public = true, file_size_limit = 5242880, allowed_mime_types = ...`, and add a migration assertion/test that catches an existing misconfigured bucket.

Oiranca (commented):
Added one follow-up review comment.

supabase/migrations/20260704000005_oir207_landing_media_bucket.sql
Oiranca: ✅ Fixed in 9fda5e57f0b123c32977fc2d75fba1993d713ce0.

Switched the `landing-media` bucket insert to `ON CONFLICT ("id") DO UPDATE SET "public" = true, "file_size_limit" = 5242880, "allowed_mime_types" = ARRAY['image/png', 'image/jpeg', 'image/webp', 'image/gif']`, so re-running the migration now forces convergence to the intended state instead of leaving a pre-existing/misconfigured bucket (e.g. `public = false`, a smaller limit, or a narrower MIME allowlist) untouched.

Also updated the existing migration sanity-check tests in `__tests__/server/uploads-service.test.ts` (describe block `migration sanity checks (OIR-207)`):
- Replaced the old "uses ON CONFLICT DO NOTHING" assertion with one that asserts `ON CONFLICT ("id") DO UPDATE SET` is used and `DO NOTHING` is not.
- Added a new test that parses the `DO UPDATE SET` clause and asserts it re-applies `public = true`, `file_size_limit = 5242880`, and the full 4-item MIME allowlist.

Note: this repo doesn't have a live-DB migration test harness (the one Supabase integration script it does have spins up a local instance via `supabase db reset`, which we intentionally avoid running here), so the added coverage is a SQL-content-level assertion consistent with the existing pattern in that file rather than a live-bucket-seed-and-reapply test.

PR #154, opened 2026-07-04
OIR-208: unified events with table-level blocks and materials (stacked on OIR-207)

Oiranca:
## Summary

- Collapses "Club events" (landing) and "Internal events" (room-booking) into a single unified `events` service and admin "Eventos" section, with a `visibleOnLanding` toggle controlling landing publication (requires `title_es`/`title_en` to be set — enforced server-side, cannot be bypassed).
- Adds room **OR per-table** event blocking: a block with a `table_id` scopes both the availability read and the reservation-cancellation predicate to that single table; a `NULL` `table_id` preserves today's whole-room blocking behavior unchanged. All read paths that consult `event_room_blocks` (`tables-service`, `rooms-service`, `reservations-service`, `saved-games-service`) were extended to respect the new column.
- Adds per-event **materials** (`event_equipment`): admin-only, service-role-gated join table with a `quantity > 0` CHECK, replaced atomically alongside room blocks via a single `apply_club_event_room_blocks` RPC call.
- Removes the legacy split admin events UI (`events-section.tsx`, `events-tab.tsx`) in favor of the unified `club-events-section.tsx`. The legacy `/api/events` route handlers remain (still admin-gated, rate-limited, and unreferenced by any UI) — kept only because existing test suites exercise them directly; documented as dead/double-write surface for a future removal pass.

## Stacked-chain note

This is the **NEW FINAL PR** of the stacked chain #148 → #149 → #150 → #151 → #152 → #153 → this PR. The user will merge **only `feat/oir-208-unified-events`** into `develop` — all prior chain PRs will auto-close as already-merged once that happens.

## Security summary

- Migration `20260704000006`: `event_room_blocks.table_id` is a nullable FK to `tables.id` with `ON DELETE CASCADE`; `event_equipment` has RLS enabled with no anon/authenticated policies (service-role only) and a `quantity > 0` CHECK constraint.
- `apply_club_event_room_blocks` RPC: old 2-arg version dropped, replaced by an unambiguous 3-arg `SECURITY DEFINER` function with a pinned `search_path`, `EXECUTE` revoked from `PUBLIC`/`anon`/`authenticated`, granted only to `service_role`.
- The table-scoped reservation-cancellation predicate only ever cancels reservations for the single table referenced by a table-scoped block — it cannot cancel beyond that table's own reservations.
- Admin privilege checks (`requireAdminSession`) are intact on every unified-service write path (`createClubEvent`, `updateClubEvent`, `deleteClubEvent`, `listAdminClubEvents`); the visibility toggle cannot publish a row to the landing without a non-empty `title_es` (and its paired `title_en`), which is validated unconditionally before every write.
- The legacy anchor-field preservation fix (65485a1) is data-preservation only — it does not relax any authorization or validation on the legacy `/api/events` write paths, which remain admin-gated + rate-limited + guarded by the `isClubEventRow` exclusion.
- Materials payloads are validated before any DB write: `quantity` must be a positive integer, and duplicate `equipmentId` entries are rejected.
- Every availability read path that gained table-granularity can only **narrow** blocking (a table-scoped block affects one table instead of the whole room) — it never widens anon/member-visible access beyond previous whole-room-blocking behavior.
- No `dangerouslySetInnerHTML`/`innerHTML`/`eval` introduced; `en.json`/`es.json` key parity verified (no missing keys either direction); no secrets or hardcoded credentials found in the diff; all commit messages are in English.

## Test plan

- ~867 tests green (2 QA rounds completed, including `__tests__/server/oir208-unified-events.test.ts` covering visibility toggle, materials, and table-granularity availability, plus regression coverage for the legacy anchor-field preservation fix).
- Post-merge: run `supabase db push` to apply `20260703*` and `20260704000001`..`20260704000006` migrations (user-only action per repo convention).


supabase/migrations/20260704000006_oir208_table_blocks_and_materials.sql
```diff
@@ -0,0 +1,168 @@
...
+      v_table_id := NULLIF(v_elem->>'table_id', '')::uuid;
+      v_date     := (v_elem->>'date')::date;
+      v_all_day  := COALESCE((v_elem->>'all_day')::boolean, false);
+      v_start    := CASE WHEN v_all_day THEN '00:00'::time
+                         ELSE (v_elem->>'start_time')::time END;
+      v_end      := CASE WHEN v_all_day THEN '23:59'::time
+                         ELSE (v_elem->>'end_time')::time END;
+
+      IF v_room_id IS NOT NULL THEN
+        INSERT INTO public.event_room_blocks (event_id, room_id, table_id, date, start_time, end_time, all_day)
```
Oiranca, line 124: A table-scoped block can pair `room_id` from one room with `table_id` from another because the schema only has independent FKs. That creates inconsistent rows and the cancellation/availability logic will act on the table id while the admin UI shows the selected room. Please enforce `table_id` belongs to `room_id` either with a composite FK/trigger in the migration or by validating in the RPC before inserting, and add a regression test for a mismatched room/table payload.

__tests__/server/oir208-unified-events.test.ts
```diff
@@ -0,0 +1,996 @@
...
+        dateKind: 'single',
+        visibleOnLanding: false,
+      })
+
+      expect(result.titleEs).toBeDefined()
+    })
+  })
+
+  describe('Materials Validation', () => {
+    it('rejects materials with quantity 0', async () => {
```
Oiranca, line 319: This test name says the service rejects quantity `0`, but the test never calls the service or asserts a thrown `ServiceError`; it only checks that a local literal is less than 1. This would still pass if `validateMaterialsPayload` stopped validating quantity entirely. Please route this payload through `createClubEvent`/`updateClubEvent` and assert a 400, same as the non-array and missing-equipment tests below.

supabase/migrations/20260704000006_oir208_table_blocks_and_materials.sql
Oiranca: ✅ Fixed in `bd73411`. Chose in-RPC validation over a composite FK (a composite FK would require adding a `UNIQUE(id, room_id)` on `tables`, a broader schema change than warranted). `apply_club_event_room_blocks` now validates, before each insert, that the block's `table_id` actually belongs to `room_id`:
```sql
IF v_table_id IS NOT NULL
   AND NOT EXISTS (SELECT 1 FROM public.tables WHERE id = v_table_id AND room_id = v_room_id)
THEN RAISE EXCEPTION 'table_id % does not belong to room_id %', v_table_id, v_room_id USING ERRCODE = '23514';
END IF;
```
`23514` (check_violation) is already mapped to a 400 `ServiceError` in `club-events-service.ts`, so no service-layer change was needed. Added a regression test routing a mismatched `roomId`/`tableId` payload through `updateClubEvent` and asserting 400. (Migration edited in place — not yet applied to any DB.)

__tests__/server/oir208-unified-events.test.ts
Oiranca: ✅ Fixed in `bd73411`. The sham quantity tests (`quantity 0` / negative) no longer assert on a local literal — they now route the materials payload through `createClubEvent`/`updateClubEvent` and assert a thrown 400 `ServiceError`, matching the adjacent non-array and missing-equipment tests. This would now fail if `validateMaterialsPayload` stopped validating quantity. oir208 suite: 33/33 passing.

Oiranca: Follow-up (post QA review, `aee85fb`): beyond this quantity-0 test, QA found 18 additional `expect(true).toBe(true)` placeholder tests in this file (54%), concentrated in the highest-risk sections (availability table-granularity, migration sanity, blocksMatchSchedules). All 18 were replaced with real assertions exercising `getTableAvailability`/`getRoomTablesAvailability`/`hasEventBlockConflict`/`assertTableAndEventAvailability` and the migration SQL — verifying the core invariant that a `table_id`-scoped block affects only that table, not siblings in the same room. File is now 34 real tests, 0 placeholders; no source bug was found.

lib/server/club-events-service.ts
```diff
@@ -0,0 +1,893 @@
...
+  const wantsBlocks = blocksRoomsProvided ? parseBooleanFlag(body.blocksRooms) : undefined
+
+  const validatedSchedules = wantsBlocks !== false && schedulesProvided
+    ? validateSchedulesPayload(body.schedules)
+    : null
+
+  const materialsProvided = body.materials !== undefined
+  const validatedMaterials = materialsProvided ? validateMaterialsPayload(body.materials) : null
+
+  const { data, error } = await admin
```
Oiranca, line 884: The unified update path persists event fields before calling the RPC that replaces blocks/materials. With OIR-208 this is broader than the earlier room-block case: a bad `tableId`, missing `equipmentId` FK, or RPC failure can return 400/500 after the event title/date/visibility fields have already changed. Please pre-validate referenced rooms/tables/equipment before this UPDATE or move the field update plus block/material replacement into one DB transaction, and cover the failure path with a test that proves no partial event update survives.

Oiranca (commented):
Review pass: previously reported OIR-208 findings were fixed; one new atomicity gap remains in the unified update path.

lib/server/club-events-service.ts
Oiranca: Fixed in two commits: c8fba2e (propagates the room-id pre-validation + compensating-revert fix from PR #149/oir-203 through the stack) and 6b4b935 (extends it to cover this PR's broader scope).

`updateClubEvent` now validates every room id, table id, and equipment id referenced in `schedules`/`materials` against the `rooms`, `tables`, and `equipment` tables (new `validateTablesExist`/`validateEquipmentExists`, alongside the existing `validateRoomsExist`) BEFORE the event-fields UPDATE runs — so a bad tableId or missing equipmentId FK is rejected with 400 and zero writes happen, not just a bad room id.

As a second layer, if `apply_club_event_room_blocks` (via `applyClubEventRoomBlocksAndMaterials`) still fails for some other reason after all references passed validation (transient DB error, etc.), the fields UPDATE is reverted back to its pre-update values — this now covers the materials-only update path too, since that RPC is invoked whenever either blocks or materials change.

Added regression tests in `__tests__/server/club-events-service.test.ts`:
- rejects an unknown table id with 400 before any UPDATE runs
- rejects a missing equipment id with 400 before any UPDATE runs
- forces the RPC to fail during a materials-only change and asserts the fields UPDATE is reverted to the original values

Full test suite for this file (78 tests) and the whole repo (885 tests) pass.

PR #159, opened 2026-07-14
feat(cron): register mark-no-show in vercel.json

Oiranca:
## Summary

Registers the `mark-no-show` cron job in `vercel.json` as part of the Supabase to Vercel migration (Pre-03). Currently this job runs via an external scheduler (cron-job.org); once the project deploys to Vercel and the external scheduler is retired, this entry ensures the job continues to run via Vercel Cron.

- **Schedule:** `*/15 * * * *` (every 15 minutes). The route performs a passive DB cleanup RPC (`markNoShowReservations`) with no real-time requirement and no documented cron-job.org cadence to match, so a conservative 15-minute interval was chosen.
- **maxDuration:** `60` seconds, to give headroom for Supabase region latency.

## Extra beyond issue scope

Also removed the stale `app/api/cron/cancel-pending` entry from the `functions` block in `vercel.json`. That route is dead code that unconditionally returns `410 Gone` (since KIM-366), so its `maxDuration` config was inert and has been dropped along with the cron registration change.

## Security

The `mark-no-show` route's `CRON_SECRET` bearer-token authentication check is present and was **not** modified by this change — `route.ts` is untouched by this PR. It rejects any request missing a valid `Authorization: Bearer <CRON_SECRET>` header (constant-time compare via `tokensMatch()`), returning `401 Unauthorized` on any missing/invalid case.

## Acceptance criteria

- [x] `vercel.json` contains a valid `crons` entry for `mark-no-show`
- [x] Route rejects requests without a valid cron auth header/secret (verified untouched, reviewed `tokensMatch()` implementation)
- [x] `pnpm build` passes
- [x] `vercel.json` is valid JSON matching Vercel's cron schema

## Test plan

- [x] JSON schema validation passed
- [x] `pnpm build` passed
- [x] 961 tests green, including 6 in `__tests__/app/api/cron/mark-no-show.test.ts`

### DIFF
--- a/app/api/cron/mark-no-show/route.ts
+++ b/app/api/cron/mark-no-show/route.ts
@@ -33,6 +33,10 @@ async function handleCronRequest(request: NextRequest) {
   }
 }
 
+export async function GET(request: NextRequest) {
+  return handleCronRequest(request)
+}
+
 export async function POST(request: NextRequest) {
   return handleCronRequest(request)
 }
--- a/vercel.json
+++ b/vercel.json
@@ -1,7 +1,13 @@
 {
   "functions": {
-    "app/api/cron/cancel-pending/route.ts": {
-      "maxDuration": 10
+    "app/api/cron/mark-no-show/route.ts": {
+      "maxDuration": 60
     }
-  }
+  },
+  "crons": [
+    {
+      "path": "/api/cron/mark-no-show",
+      "schedule": "*/15 * * * *"
+    }
+  ]
 }
--- a/.claude/agent-progress.md
+++ b/.claude/agent-progress.md
@@ -833,3 +833,9 @@ Real-time log of all agent work. Agents append entries as work progresses.
 - [11:26] Security Status: APPROVE — 0 CRITICAL/HIGH/MEDIUM, 0 LOW findings
 - [11:26] ✅ Complete — opening PR chore/pre-push-full-validation -> main
 - [11:26] PR opened: https://github.com/KimoxStudio/alea-webapp/pull/157 (chore/pre-push-full-validation -> main)
+
+#### [fe9fedd6] software-engineer -- register mark-no-show cron
+- [16:48] Started; confirmed CRON_SECRET auth already present in app/api/cron/mark-no-show/route.ts (untouched)
+- [16:48] Verified mark_no_show_reservations DB function is a passive cleanup query (marks reservations no_show once end_time has passed); no strict timing docs found, chose */15 * * * * cadence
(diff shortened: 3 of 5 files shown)


vercel.json
```diff
@@ -1,7 +1,13 @@
...
-    "app/api/cron/cancel-pending/route.ts": {
-      "maxDuration": 10
+    "app/api/cron/mark-no-show/route.ts": {
+      "maxDuration": 60
     }
-  }
+  },
+  "crons": [
+    {
+      "path": "/api/cron/mark-no-show",
```
Oiranca, line 9: Vercel Cron invokes the configured path with HTTP GET, but `app/api/cron/mark-no-show/route.ts` currently exports only `POST`. This means the scheduled job will hit `/api/cron/mark-no-show` without executing `handleCronRequest`, so no-show reservations will not be marked in production. Please add a GET handler that shares the same CRON_SECRET validation path, and add/adjust a route test that exercises the Vercel Cron GET invocation. Reference: https://vercel.com/docs/cron-jobs#how-cron-jobs-work

app/api/cron/mark-no-show/route.ts
```diff
@@ -33,6 +33,10 @@ async function handleCronRequest(request: NextRequest) {
   }
 }
```
Oiranca, line 35: The GET export fixes the production behavior, but the PR still has no route test proving the Vercel Cron path works through GET with the same CRON_SECRET checks. Please add coverage for at least: valid GET returns the mark count, missing/invalid bearer returns 401, and POST remains supported if you want to keep the manual trigger. Without this, a future refactor can break the scheduled path while the existing POST tests stay green.

vercel.json
Oiranca: ✅ Fixed in commit edddd1e — added `export async function GET(request: NextRequest) { return handleCronRequest(request) }` to `app/api/cron/mark-no-show/route.ts`, reusing the same `CRON_SECRET`/`tokensMatch` bearer-token validation already used by `POST`. No new auth logic was written. `POST` remains supported for manual/local triggering. Merged to main in PR #159.

app/api/cron/mark-no-show/route.ts
Oiranca: ✅ Test coverage added in commit acd3b44 — 6 new GET test cases in `__tests__/app/api/cron/mark-no-show.test.ts` mirroring the existing POST suite: valid GET returns the mark count (200), missing/invalid bearer returns 401, missing CRON_SECRET env returns 401, service error returns 500, zero-count case returns 200. All existing POST tests remain green (12/12 total, full suite 967/967). Merged to main in PR #159.

PR #162, opened 2026-07-14
Pre-02: verify serverless-safe rate limiter (already resolved via Upstash Redis) + strengthen persistence test

Oiranca:
## Summary

This migration blocker (Pre-02) was found **already resolved on `main`**: `lib/server/security.ts` uses `@upstash/ratelimit` + `@upstash/redis` (sliding-window, external shared store) whenever `UPSTASH_REDIS_REST_URL` / `UPSTASH_REDIS_REST_TOKEN` are set, replacing the old `globalThis` in-memory `Map` that did not survive across Vercel serverless isolates. This was implemented as part of KIM-401 (rate-limit hardening), prior to this PR.

This PR does **not** change any production code. It:
- Adds the Pre-02 spec doc at `docs/issues/migration-pre-02-serverless-rate-limiter.md`, documenting the blocker and the resolution.
- Adds one new test in `__tests__/server/security.test.ts` — "shares rate-limit counter state across multiple requests (stateful persistence)" — that proves a shared/persisted counter is enforced across sequential requests via the Redis-backed path. The pre-existing Redis test only proved client/`Ratelimit` instance memoization (constructor called once, `limit()` called per request); it did not prove the counter itself is shared/stateful. The new test closes that acceptance-criterion gap explicitly.

## Decision flagged for your awareness

The issue spec offered two options for the storage mechanism:
1. A dedicated Postgres table (preferred — works today against Supabase and later against Neon, no new dependency).
2. An external store (e.g. Upstash Redis) — flagged in the original issue as an open question because it adds a new dependency.

The implementation on `main` took **option 2 (Upstash Redis)**, not option 1. `@upstash/ratelimit` and `@upstash/redis` are already present in `package.json` on `main` (not added by this PR). This works correctly across serverless isolates and will continue to work after the eventual Neon migration (it's independent of the Postgres/Supabase choice), but it is an already-decided architectural choice worth your explicit sign-off since it wasn't the option marked "preferred" in the original spec.

## Verification

- `enforceRateLimit(request, policy)` signature is unchanged — no call-site changes needed.
- `pnpm build` and `pnpm typecheck` pass.
- `pnpm vitest run __tests__/server/security.test.ts` — 16/16 pass (15 pre-existing + 1 new).
- Diff is limited to `__tests__/server/security.test.ts` and `docs/issues/migration-pre-02-serverless-rate-limiter.md` — no production/source file touched, no `package.json` change.

## Security review notes (non-blocking, pre-existing implementation — informational only)

- **Fail-open/closed on Redis error**: `enforceRateLimitRedis` calls `ratelimit.limit(identifier)` with no local try/catch; if the Upstash call throws (timeout, outage, auth failure), the rejection propagates to the caller of `enforceRateLimit` rather than being handled explicitly in `security.ts`. Worth confirming call sites don't silently swallow this in a way that either fails the request outright (fail-closed, availability risk) or allows the request through uninspected (fail-open, rate-limit bypass risk) without an explicit decision either way.
- **Client identifier derivation**: `getClientAddress()` only trusts `x-forwarded-for` when `TRUST_PROXY_HEADERS=true` AND the `x-real-ip` source is in the configured trusted-proxy CIDR list (default: loopback only); otherwise it falls back to `x-real-ip` directly, which is reasonable but note `x-real-ip` itself is not intrinsically more trustworthy than `x-forwarded-for` unless the deployment platform (e.g. Vercel) is confirmed to set/overwrite it at the edge before the app sees it — worth a quick confirmation against Vercel's actual header behavior in production, outside the scope of this PR.

These are existing-code observations, not introduced by this PR, and do not block merge.

Ref: Pre-02 (Supabase→Neon migration blocker, Pre phase).


__tests__/server/security.test.ts
```diff
@@ -440,5 +440,58 @@ describe('server security helpers', () => {
       expect(mockRatelimitConstructor).toHaveBeenCalledTimes(1)
       expect(mockLimit).toHaveBeenCalledTimes(2)
     })
+
```
Oiranca, line 443: This test proves that the local mock counter works, but it does not exercise the serverless failure mode described in the issue. `sharedCounter` lives in the test process and the module is imported only once, so `_redisClient` / `_ratelimitCache` are never reset like they would be across cold starts. Please make the regression cover the actual contract: import/use `enforceRateLimit`, reset modules to simulate a fresh serverless instance, re-import it, then verify the second instance still observes the externally persisted Upstash state. Otherwise this test can pass even if the production implementation regresses back to process-local memory.

Oiranca: ✅ Fixed in cfed1bf.

You're right — the original test only proved that a locally-scoped `sharedCounter` `Map` object persists across sequential calls within a single Vitest module import, which never touches the actual claim of this migration (that state survives a serverless cold start because it lives in Upstash, not process memory).

Rewrote the test (`persists rate-limit counter state across a simulated serverless cold start`) to:
1. Call `vi.resetModules()` and dynamically re-`import('@/lib/server/security')` to get a first fresh module instance ("Instance A"), simulating a cold start. Its module-level `_redisClient` / `_ratelimitCache` singletons are new/empty.
2. Drive it to the rate limit via the shared `mockLimit` implementation (backed by a counter map that lives in the *test file's* scope — standing in for the real Upstash Redis store, since `vi.resetModules()` does not clear that).
3. Call `vi.resetModules()` again and re-import a second, fully independent module instance ("Instance B") — asserting `instanceB.enforceRateLimit !== instanceA.enforceRateLimit` to confirm it's genuinely a distinct module, not a cached reference.
4. Assert that Instance B's very first request is already rejected (429), proving the persisted state was observed by the *fresh* module rather than reset with it.

I verified this actually catches the regression it targets: I temporarily patched `enforceRateLimitRedis` to count requests in a local `Map` instead of delegating to `ratelimit.limit()`, ran the test, confirmed it failed (`expected 429, got undefined`), then reverted the change. `lib/server/security.ts` itself needed no changes — this was purely a test-coverage gap, as you diagnosed.

Full validation: `pnpm typecheck`, `pnpm test` (64 files / 962 tests), and `pnpm build` all pass on the updated branch.

PR #163, opened 2026-07-14
feat(db): introduce lib/db seam for Postgres access (F0-05)

Oiranca:
## Summary

Introduces `lib/db/index.ts` as the single indirection seam for Postgres access, exposing `getDb()` (user-scoped, RLS-respecting) and `getAdminDb()` (admin, RLS-bypassing) as thin wrappers around today's `createSupabaseServerClient()` / `createSupabaseServerAdminClient()` in `lib/supabase/server.ts`. Migrates all 14 `lib/server/*.ts` call sites to import from `lib/db` instead of importing Supabase directly.

This is phase F0 ("abstraction seams") of the Supabase→Vercel/Neon migration plan — issue #5. Pure indirection, no behavior change; prep for a later Drizzle/Neon swap (F1).

- Spec: `docs/issues/migration-f0-05-lib-db-seam.md`
- Plan: `docs/MIGRATION-supabase-to-neon.md`, issue #5

## Note on diff scope

This branch's history includes the already-merged `migration-pre-01..04` commits from `main` (PRs #155–#162), which `develop` has not yet absorbed independently. The commits relevant to **this** PR are only:

- `5302bad` — feat(db): introduce lib/db seam for Postgres access (F0-05)
- `8179c64` — test(db): add smoke test for lib/db seam (F0-05)

Everything else in the diff (vercel.json, package.json, docs, scripts, unrelated tests) is pre-existing work already on `main`, not new in this PR. Recommend a separate `develop`⟵`main` sync (user-only merge decision) to close that gap independently of this PR.

## Changes

- New `lib/db/index.ts`: `getDb()` / `getAdminDb()`, documents the user-scoped vs admin RLS convention.
- 14 `lib/server/*.ts` files migrated to import from `lib/db` instead of `lib/supabase/server`. `lib/supabase/server.ts` itself is unchanged (still the underlying factory).
- `__tests__/lib/db.test.ts` (added by qa-engineer): smoke test locking in that `getDb()`/`getAdminDb()` route to the correct factory and return distinct clients.

## Review performed (security-reviewer)

- Diffed every call-site change; confirmed a strict 1:1, no-swap mapping (`createSupabaseServerClient` → `getDb()`, `createSupabaseServerAdminClient` → `getAdminDb()`) across all 14 files — no accidental admin/user-scoped client swap.
- Confirmed `lib/supabase/server.ts` is untouched (still the sole underlying factory).
[...]

PR #164, opened 2026-07-15
feat(auth): introduce lib/auth/session seam for Supabase Auth (F0-06)

Oiranca:
## Summary

Introduces `lib/auth/session` — a thin seam wrapping today's Supabase Auth method calls (`.auth.getUser()`, `.auth.signInWithPassword()`, `.auth.signOut()`, `.auth.admin.createUser()`, `.auth.admin.deleteUser()`, `.auth.admin.updateUserById()`) behind named, stable functions. This is an F0 migration-prep step (see `docs/MIGRATION-supabase-to-neon.md`) so a later phase can swap the underlying auth implementation (Supabase Auth -> Auth.js) without touching call sites again. This mirrors the pattern established in PR #163 (F0-05, `lib/db` seam).

**This is a pure refactor with no intended behavior change.** No Auth.js implementation is introduced in this PR.

New exported functions in `lib/auth/session/index.ts`:
- `getAuthUser(client)` — resolves the current Supabase Auth user, returns `null` on error or no session (same "no session vs auth error" non-distinction as the prior inline code)
- `signInWithPassword(client, credentials)` — passthrough to `client.auth.signInWithPassword`
- `signOut(client)` — passthrough to `client.auth.signOut`
- `createAuthUser(admin, input)` — admin-only, passthrough to `admin.auth.admin.createUser`
- `deleteAuthUser(admin, id)` — admin-only, passthrough to `admin.auth.admin.deleteUser`
- `updateAuthUserById(admin, id, attributes)` — admin-only, passthrough to `admin.auth.admin.updateUserById`

### Files changed
- `lib/auth/session/index.ts` (new)
- `lib/server/auth.ts` — `getSessionUser` now uses `getAuthUser`
- `lib/server/auth-service.ts` — `login`, `register`, `logout`, `logoutWithClient`, `activateAccount`, `recoverAccount` now use the seam
- `lib/server/users-service.ts` — member import and `updateUser`/`deleteUser` admin flows now use the seam

## Validation

- QA (qa-engineer) approved: typecheck/build green, behavior-parity and completeness verified against the original inline `.auth.*` calls.
- Security review (this PR): confirmed all admin-only functions (`createAuthUser`/`deleteAuthUser`/`updateAuthUserById`) still require an explicit admin Supabase client argument supplied by the caller (no default/implicit client, not reachable from user-scoped code paths), no swallowed auth errors introduced, no secrets/credentials added, no new attack surface, and scope stayed within the 4 files above (`lib/server/tables-service.ts` and `lib/server/uploads-service.ts` untouched, no test files added).
- Local CI (pre-push hook) also ran clean: typecheck, lint, 968 tests passed, and production build succeeded.


lib/auth/session/index.ts
```diff
@@ -0,0 +1,152 @@
...
+}
+
+/** Updates a Supabase Auth user's attributes (email, password, email_confirm) (admin-only). */
+export async function updateAuthUserById(
+  admin: AuthAdminClient,
+  id: string,
+  attributes: UpdateAuthUserAttributes,
+): Promise<AuthAdminOperationResult> {
+  return admin.auth.admin.updateUserById(id, attributes)
+}
```
Oiranca, line 152: Please add a focused unit test for this new auth seam before merging. This module now owns the indirection for session reads, password sign-in/sign-out, and admin user management, including security-sensitive calls such as `createAuthUser`, `deleteAuthUser`, and `updateAuthUserById`. The sibling seams in PR #163 and #165 have direct smoke tests that lock the wrapper-to-underlying-client mapping; this one should do the same so a future refactor cannot accidentally route an admin operation through the wrong method, drop credentials, or change the existing auth-error semantics without a failing test.

Oiranca (commented):
I found one test-coverage issue. The refactor introduces a security-sensitive auth seam, but it does not add direct coverage for the new wrapper API.

lib/auth/session/index.ts
Oiranca: ✅ Added a focused unit test for this seam: `__tests__/lib/auth/session.test.ts` (11 tests, following the same conventions as the sibling F0-05 (`lib/db`) and F0-07 (`lib/storage/qr`) seam tests).

Since `lib/auth/session` doesn't wrap the client *factories* — call sites pass their own client in — each test builds a minimal mock client per function rather than mocking `lib/supabase/server`, and asserts the wrapper routes to the correct underlying `.auth` / `.auth.admin` method with the right arguments and preserves the result/error shape unchanged:

- `getAuthUser()` — resolved user, Supabase error → `null`, no user → `null`
- `signInWithPassword()` — success + error semantics preserved
- `signOut()` — success + error propagated
- `createAuthUser()`, `deleteAuthUser()`, `updateAuthUserById()` — correct admin method + args, results/errors returned unchanged

Full suite: 65 test files / 972 tests passing. `pnpm typecheck`, `pnpm lint`, and `pnpm build` all clean. Pushed to `migration-f0-06-auth-session-seam` (commit 0cffe9e).

PR #165, opened 2026-07-15
feat(storage): introduce lib/storage/qr seam for Supabase Storage (F0-07)

Oiranca:
## Summary

Introduces `lib/storage/qr/index.ts` as a single seam for Supabase Storage access, per the F0 migration plan (see `docs/MIGRATION-supabase-to-neon.md`). This is a pure indirection layer with **zero behavior change** — Supabase Storage remains the actual backend for now; the seam only exists so a later phase (F1+) can swap the storage backend (e.g. to Vercel Blob) without touching call sites again.

- New module exports `uploadToStorage`, `getPublicStorageUrl`, `removeFromStorage`, wrapping `createSupabaseServerAdminClient().storage`.
- `lib/server/tables-service.ts` (`table-qr-codes` bucket, QR code generation) and `lib/server/uploads-service.ts` (`landing-media` bucket, admin-uploaded images) migrated off direct `admin.storage.*` calls onto this seam.
- No changes to bucket names, access scope (still admin/service-role client only), privilege checks, or path construction — confirmed via diff review.
- `lib/server/auth-service.ts` / `lib/server/auth.ts` untouched (belongs to sibling issue F0-06 on a separate branch).

## Files changed
- `lib/storage/qr/index.ts` (new)
- `lib/server/tables-service.ts`
- `lib/server/uploads-service.ts`
- `__tests__/lib/storage/qr.test.ts` (new, added by qa-engineer)
- `.claude/agent-progress.md` (log reconciliation)

## Test coverage
qa-engineer added 7 tests in `__tests__/lib/storage/qr.test.ts` covering `uploadToStorage` (call args, option preservation, default `upsert`, error wrapping), `getPublicStorageUrl` (URL extraction, null handling), and `removeFromStorage` (call args, error wrapping) against a mocked admin client. Full suite: 975/975 tests passing. Typecheck, lint, and build all green (verified locally via pre-push hook and independently by security-reviewer).

## Security review
Reviewed by security-reviewer: no bucket/scope changes, no secrets/credentials exposed, no remaining direct `.storage.` usage outside the seam (verified via grep), no scope creep (auth files untouched, no unrelated test files). APPROVED.


lib/storage/qr/index.ts
```diff
@@ -0,0 +1,88 @@
...
+  body: Buffer | Uint8Array,
+  options: StorageUploadOptions = {},
+): Promise<StorageOperationResult> {
+  const storage = getAdminStorage()
+  const { error } = await storage.from(bucket).upload(path, body, {
+    contentType: options.contentType,
+    upsert: options.upsert ?? false,
+  })
+
+  return { error: error ? { message: error.message } : null }
```
Oiranca: Please avoid narrowing the Supabase Storage error down to only `{ message }` here. `uploadLandingMediaImage()` explicitly says the underlying storage failure should be logged server-side for diagnostics, but after this wrapper the caller only receives `error.message`; useful fields such as `name`, `statusCode`, provider-specific codes, or the original error object are lost. Keep the public client response generic, but preserve the full error object (or at least structured diagnostic fields) inside the server-side seam so bucket misconfiguration, permissions, quota, and backend outages remain distinguishable in logs and tests.

Oiranca (commented):
I found one observability regression in the Storage seam.

lib/storage/qr/index.ts
Oiranca: ✅ Fixed. `StorageOperationResult.error` in `lib/storage/qr/index.ts` is now a structured `StorageErrorDetail` (`name`, `status`, `statusCode` alongside `message`) instead of being narrowed to `{ message }`. `uploadLandingMediaImage()` in `lib/server/uploads-service.ts` now logs the full structured error detail server-side rather than just `error.message`, so bucket misconfiguration, permissions, quota, and backend outages remain distinguishable in logs. Covered by new test cases in `__tests__/lib/storage/qr.test.ts`. See commit c162d06.

PR #167, opened 2026-07-16
chore: group lib/server/ by domain (KIM-412)

Oiranca:
## Summary
- Regroups `lib/server/` from a flat 24-file structure into domain subdirectories: `auth/`, `equipment/`, `events/`, `games/`, `partners/`, `reservations/`, `rooms/`, `shared/`, `tables/`, `uploads/`, `users/`
- Rewrites import paths across `app/`, `lib/`, and `__tests__/` (~120 files) to match the new locations
- Fixes `middleware.ts` relative import for the moved `security-edge` module
- Updates `vitest.config.mts` per-file coverage threshold keys to the new paths

Pure file-move refactor — no behavior changes.

## Test plan
- [x] `pnpm typecheck` — clean
- [x] `pnpm lint` — no warnings/errors
- [x] `pnpm test --run` — 990/990 tests passed (67 files)
- [x] `pnpm build` — production build succeeds

Ref: KIM-412


Oiranca (commented):
Reviewed the refactor with focus on best practices, security boundaries, and test coverage. The TypeScript imports are consistent and the PR validates locally: `pnpm typecheck`, `pnpm lint`, `pnpm test --run` (990 tests), and `pnpm build` all pass in a temporary worktree. Note: local Node is `26.4.0`, outside the repo engine range `>=20.11.1 <25`, so CI on the supported Node version is still the authoritative result.

One broader follow-up: several repo docs still point at the old server paths after this regrouping (`CLAUDE.md`, `docs/ARCHITECTURE.md`, `docs/SECRET-ROTATION-CHECKLIST.md`, and migration docs). Please update at least the security/auth architecture references in this PR or a clearly linked follow-up, because stale security-path docs make future imports and reviews easier to get wrong.

lib/server/shared/security.ts
```diff
@@ -5,7 +5,7 @@ import type { Redis } from '@upstash/redis'
 import type { Ratelimit } from '@upstash/ratelimit'
 
 // Re-export all Edge-safe helpers so that existing route-handler imports from
-// `@/lib/server/security` continue to work without any call-site changes.
+// `@/lib/server/shared/security` continue to work without any call-site changes.
```
Oiranca: Please update this comment while moving the file. The old `@/lib/server/security` module no longer exists, and this PR did require call-site rewrites, so the “existing route-handler imports ... without any call-site changes” framing is now misleading. For a security module, the comment should state the current boundary directly: route handlers may import the Node-only `shared/security`, while middleware must import the Edge-safe `shared/security-edge`.

lib/server/shared/security-edge.ts
```diff
@@ -10,7 +10,7 @@
  *
  * Node-only helpers (`tokensMatch`, `enforceMutationSecurity`, rate limiting)
  * stay in `security.ts`, which re-exports everything from this file so that
- * existing route-handler imports from `@/lib/server/security` are unaffected.
+ * existing route-handler imports from `@/lib/server/shared/security` are unaffected.
```
Oiranca: Please tighten this comment as part of the security refactor. `shared/security.ts` imports Node `crypto`, while this file is the Edge-safe boundary used by middleware. Saying route-handler imports are “unaffected” after the path move is ambiguous; future reviewers should be able to tell from this block that `shared/security-edge` is the only middleware-safe module and `shared/security` is intentionally Node-only.

lib/server/shared/security.ts
Oiranca: ✅ Fixed in 976e773.

Reworded the comment to state the current boundary directly instead of the stale "unaffected import" framing:

> Re-export all Edge-safe helpers to keep the Node-only vs Edge-safe boundary explicit: route handlers should import the Node-only helpers below from `@/lib/server/shared/security`, while middleware and other Edge runtime code must import `@/lib/server/shared/security-edge` directly, which is Node `crypto`-free.

lib/server/shared/security-edge.ts
Oiranca: ✅ Fixed in 976e773.

Reworded to make the Node-only vs Edge-safe boundary explicit and unambiguous (kept consistent with the wording used in the companion comment on `security.ts`):

> Node-only helpers (`tokensMatch`, `enforceMutationSecurity`, rate limiting) stay in `security.ts`, which re-exports everything from this file. The boundary is explicit: route handlers should import the Node-only helpers from `@/lib/server/shared/security`, while middleware and other Edge runtime code must import `@/lib/server/shared/security-edge` directly — `security.ts` is intentionally Node-only and is not Edge-safe.

PR #170, opened 2026-07-16
feat(F1): scaffold Auth.js Credentials provider with bcrypt (KIM-416)

Oiranca:
## Summary

Adds a new, inert, parallel Auth.js (NextAuth v5) Credentials + bcrypt scaffold as part of the F1 phase of the Supabase → Vercel/Neon migration. This is infrastructure-only: it does not touch, replace, or modify the existing Supabase auth flow. No existing route, page, or session logic is changed.

- New Auth.js v5 config (`lib/authjs/*`) with a Credentials provider backed by bcrypt password verification
- New route handler at `app/api/authjs/[...nextauth]`
- Fully isolated from the current Supabase-based auth (`lib/server/auth*`, `app/api/auth/*`) — both systems coexist without interference

## Test plan

- `pnpm typecheck` — passes
- `pnpm lint` — passes (no ESLint warnings or errors)
- `pnpm exec vitest run` — 1013/1013 tests passing across 69 files, including new suites:
  - `__tests__/authjs/credentials-user.test.ts` (14 tests)
  - `__tests__/authjs/config.test.ts` (9 tests)
- `pnpm build` — production build succeeds; all routes compile, including `/api/authjs/[...nextauth]`

Branch was rebased onto current `develop` (post PR #167) prior to validation; all four checks (typecheck, lint, test, build) re-confirmed via the repo's pre-push hook on force-push.


lib/authjs/credentials-user.ts
```diff
@@ -0,0 +1,71 @@
...
+): Promise<AuthJsUser | null> {
+  const pool = getAuthDbPool()
+
+  if (!pool) {
+    return null
+  }
+
+  try {
+    const result = await pool.query<ProfileRow>(
+      'SELECT id, email, password_hash, full_name FROM profiles WHERE email = $1 LIMIT 1',
```
Oiranca, line 57: This query depends on `profiles.password_hash`, but PR #169's Drizzle schema for `profiles` does not define that column. Because this function catches every DB error and returns `null`, the integration failure will look exactly like bad credentials and can slip through unit tests that mock the pool. Please either add the column in the schema PR before merging this, or gate this PR with an integration/smoke test that runs `verifyCredentials()` against the generated Drizzle schema so missing columns fail loudly.

app/api/authjs/[...nextauth]/route.ts
```diff
@@ -0,0 +1,16 @@
...
+ * which is reserved for the existing, live Supabase Auth (GoTrue) flow.
+ * Nothing in the app links to or calls this route yet — it is inert until
+ * a future cutover issue (KIM-419/420) wires it in.
+ *
+ * Forced to the Node.js runtime because the Credentials provider talks to
+ * Postgres via `pg`, which is not Edge-compatible.
+ */
+export const runtime = 'nodejs'
+
+export const { GET, POST } = handlers
```
Comment (Oiranca) [inline app/api/authjs/[...nextauth]/route.ts]: Mounting `POST` here makes the Credentials endpoint publicly reachable as soon as the route ships; it is not truly inert just because no UI links to it. The existing Supabase login route applies CSRF/mutation checks and `RATE_LIMIT_POLICIES.authLogin`, but this Auth.js route bypasses the repo's rate-limit seam. Please add an explicit feature flag/404 guard until cutover, or wrap the credentials callback/route with the same login throttling so this does not become an unadvertised password-guessing surface when `POSTGRES_URL` and `AUTH_SECRET` are present.

Oiranca (commented):
Reviewed with focus on auth security and integration with the parallel Drizzle schema. I found two issues to address before this is ready for cutover use.

lib/authjs/credentials-user.ts
Oiranca: Confirmed — verified against `origin/migration-f1-drizzle-schema` (PR #169): `lib/db/schema/profiles.ts` indeed has no `password_hash` column. This is real, but it's a known, intentionally-deferred gap rather than an oversight: Supabase never had this column either (password hashes live in Supabase-managed `auth.users.encrypted_password` today), and per `docs/MIGRATION-supabase-to-neon.md` §F2, `profiles.password_hash` is only *added and populated* atomically during the F2 cutover migration — adding it piecemeal in the F1 schema PR (#169) would create a column with no data source until then.

✅ Fixed the "fails silently" concern the other way: expanded the JSDoc on `verifyCredentials()` to explicitly document this gap and its resolution timeline, and — see my reply on the other thread — the route this function backs is now gated 404-by-default behind a new `AUTH_JS_ENABLED` flag, so this code path cannot execute in any deployed environment until the F2 cutover intentionally flips it on. Adding the actual column belongs to #169/KIM-417 (or a dedicated F2 migration task), and an integration/smoke test against a live Drizzle-generated schema is QA's call for that future work — flagging both to the team rather than doing them here, since they're out of this PR's (KIM-416, F1 Auth.js scaffolding) scope.

app/api/authjs/[...nextauth]/route.ts
Comment (Oiranca) [reply app/api/authjs/[...nextauth]/route.ts]: Good catch — you're right that "nothing links to it yet" doesn't mean it's actually unreachable once deployed with `AUTH_SECRET`/`POSTGRES_URL` set, and unlike `/api/auth/login` this route had no rate-limit seam.

✅ Fixed: `app/api/authjs/[...nextauth]/route.ts` now wraps `handlers.GET`/`handlers.POST` behind an explicit `AUTH_JS_ENABLED` check (`.env.example` documents it, defaulting to unset/false) — the route returns `404` for every request unless that flag is explicitly set to `"true"`. This keeps the endpoint genuinely inert regardless of which other Auth.js env vars are present, closing the gap before the KIM-419/420 cutover work wires this route into the same `enforceMutationSecurity`/`RATE_LIMIT_POLICIES.authLogin` seam the existing Supabase login route uses. I kept the fix to a feature-flag guard rather than wiring in the real rate limiter here, since actually activating throttling/session behavior on this route is cutover work (KIM-419/420), out of scope for this F1 scaffolding PR (KIM-416) per the task brief.

lib/authjs/credentials-user.ts
Oiranca: This is now stale after PR #169's update: `profiles.password_hash` exists in the Drizzle schema and `0000_fine_magma.sql`. Please update this JSDoc so it no longer says the column does not exist until F2. The important warning is now different: this provider depends on PR #169's schema column and remains unreachable until `AUTH_JS_ENABLED=true`.

app/api/authjs/[...nextauth]/route.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Guarded behind `AUTH_JS_ENABLED` so this endpoint 404s by default even if
+ * `AUTH_SECRET`/`POSTGRES_URL` are already present in an environment ahead of
+ * cutover — it must not become a publicly reachable, unrate-limited
+ * password-guessing surface before KIM-419/420 wires it in behind the same
+ * throttling as `/api/auth/login`.
+ */
+export const runtime = 'nodejs'
+
+function isAuthJsEnabled(): boolean {
```
Comment (Oiranca) [inline app/api/authjs/[...nextauth]/route.ts:23]: Please add route tests for the new `AUTH_JS_ENABLED` guard: `GET` and `POST` should return 404 when the env var is unset/false, and delegate to `handlers.GET`/`handlers.POST` only when it is exactly `true`. This flag is now the security control that keeps the credentials endpoint inert before cutover, so it should not remain untested.

Oiranca (commented):
Follow-up review after the latest push. The feature-flag guard addresses the exposure risk, but it needs direct test coverage because it is now the security control keeping the route inert.

lib/authjs/credentials-user.ts
Oiranca: ✅ Fixed in eaa6c68 — the JSDoc in `verifyCredentials()` no longer claims `profiles.password_hash` doesn't exist. Verified against `origin/migration-f1-drizzle-schema` (PR #169): the column is present in both `lib/db/schema/profiles.ts` and `lib/db/migrations/0000_fine_magma.sql`. Updated the comment to state that the column now exists but remains **unpopulated** until the F2 cutover migration (KIM-419) copies bcrypt hashes over from `auth.users.encrypted_password` — so every row's `password_hash` is still `null` in practice, and this function still returns `null` for every lookup today. The security posture is unchanged, only the framing of *why* was corrected.

app/api/authjs/[...nextauth]/route.ts
Comment (Oiranca) [reply app/api/authjs/[...nextauth]/route.ts:23]: Agreed this needs coverage since `AUTH_JS_ENABLED` is now the security control keeping this endpoint inert. Not implementing the tests in this reply: per this repo's convention, test files (`__tests__/**`) are exclusively owned by the `qa-engineer` agent in our pipeline — `software-engineer`/PR-response work never creates or modifies test files, to keep test authorship consistent and reviewable. Flagging this as a qa-engineer follow-up task: add route tests for `app/api/authjs/[...nextauth]/route.ts` covering (1) `GET`/`POST` return 404 when `AUTH_JS_ENABLED` is unset/false, and (2) both delegate to `handlers.GET`/`handlers.POST` when it is exactly `'true'`. Will track this to make sure it lands before the PR merges.

Comment (Oiranca) [reply app/api/authjs/[...nextauth]/route.ts:23]: ✅ Route tests added and passing.

Added: `__tests__/app/api/authjs-route.test.ts` with 5 tests covering:
- GET/POST return 404 when AUTH_JS_ENABLED is unset, false, or any value other than 'true'
- GET/POST delegate to handlers.GET/handlers.POST when AUTH_JS_ENABLED is exactly 'true'
- Mocks auth handlers and validates that the correct handler is called with the request

All tests pass. Tests follow the existing repo conventions for route handler test patterns.

PR #176, opened 2026-07-25
feat(storage): wire lib/storage/qr seam to Vercel Blob adapter (KIM-431)

Oiranca:
## Summary

- Wires `lib/storage/qr/index.ts` — the single storage seam used by both admin QR code generation and admin landing-media uploads — to delegate to the Vercel Blob adapter (`lib/storage/qr/vercel-blob.ts`) instead of Supabase Storage. All `@supabase/*` imports are removed from the seam. Function signatures (`uploadToStorage`, `getPublicStorageUrl`, `removeFromStorage`) and the `StorageErrorDetail` structured error shape are unchanged, so `lib/server/uploads/uploads-service.ts` required no edits.
- The Vercel Blob adapter itself (`lib/storage/qr/vercel-blob.ts`) was originally built — but intentionally left unwired — in KIM-421 / PR #174. This PR is the actual F3 cutover: it activates that adapter as the sole active storage backend.
- Adds an application-level 5MB upload size cap in the seam (`uploadToStorage`), since Vercel Blob has no bucket-level `file_size_limit` equivalent to what Supabase Storage enforced. Oversized bodies are rejected before the adapter is ever called, returning a structured `StorageErrorDetail` (`name: 'StoragePayloadTooLargeError'`, `status: 413`, `statusCode: '413'`).
- Adds `BLOB_READ_WRITE_TOKEN` and `BLOB_PUBLIC_BASE_URL` to `.env.example` (blank placeholders, no real values).

## Scope note: `lib/server/tables/tables-service.ts`

`uploadQrCodeToStorage()` is changed to resolve the QR code's public URL via the seam's `getPublicStorageUrl('table-qr-codes', storagePath)` instead of manually building a Supabase public URL from `NEXT_PUBLIC_SUPABASE_URL`. This is intentionally in scope for KIM-431 (already reflected in the issue's Linear scope) — without this change, QR uploads would successfully write to Vercel Blob but return a stale/dead Supabase-shaped URL, since the old manual URL construction assumed the Supabase bucket layout. This was in fact a documented required follow-up left in `vercel-blob.ts`'s original doc comment from KIM-421/PR #174, and is resolved here as part of the actual cutover.

The new code path throws via `serviceError()` (which is typed `never`) if `getPublicStorageUrl()` resolves to `null`, so there is no null-URL fallthrough — this matches the existing error-handling pattern in the same function.

## Test changes (qa-engineer)

- `tests/unit/lib/storage/qr.test.ts` rewritten for Blob-routed behavior (21/21 passing), including new tests for the 5MB cap boundary (exactly 5MB allowed, 5MB + 1 byte rejected with the structured 413 error).
- `tests/unit/server/tables-service.test.ts` and `tests/unit/server/uploads-service.test.ts` updated with `@vercel/blob` mocks and Blob-URL assertions in place of Supabase-URL ones. Verified as a genuine regression fix, not a workaround: these same test files pass on clean `develop` (20/20 and 32/32 respectively), confirming their prior failures on this branch were caused by the storage backend swap, not pre-existing flake.

## Validation

- Full suite: 1142 passing / 0 failing / 21 skipped
- `pnpm typecheck`: pass
- `pnpm lint`: pass
- `pnpm build`: pass

## References

- Linear: KIM-431
- Related: KIM-421 / PR #174 (built the Vercel Blob adapter this PR wires up)
- Migration umbrella: KIM-393..422 (Supabase → Vercel migration)

### DIFF
--- a/tests/unit/server/uploads-service.test.ts
+++ b/tests/unit/server/uploads-service.test.ts
@@ -168,6 +168,13 @@ describe('uploads-service', () => {
       expect(result.url).toContain('https://example.public.blob.vercel-storage.com')
       expect(result.url).toContain('landing-media')
       expect(result.url).toMatch(/events\/[0-9a-f-]+\.png$/)
+
+      // Verify @vercel/blob put() was called with correct arguments
+      const blobPut = vi.mocked(await import('@vercel/blob')).put
+      expect(blobPut).toHaveBeenCalled()
+      const [pathname, body, options] = blobPut.mock.calls[0]
+      expect(pathname).toMatch(/^landing-media\/events\/[0-9a-f-]+\.png$/)
+      expect(options?.contentType).toBe('image/png')
     })
 
     it('admin uploads valid JPEG file → extension .jpg derived from MIME', async () => {
@@ -188,6 +195,13 @@ describe('uploads-service', () => {
 
       expect(result).toHaveProperty('url')
       expect(result.url).toMatch(/partners\/[0-9a-f-]+\.jpg$/)
+
+      // Verify @vercel/blob put() was called with correct arguments
+      const blobPut = vi.mocked(await import('@vercel/blob')).put
+      expect(blobPut).toHaveBeenCalled()
+      const [pathname, body, options] = blobPut.mock.calls[0]
+      expect(pathname).toMatch(/^landing-media\/partners\/[0-9a-f-]+\.jpg$/)
+      expect(options?.contentType).toBe('image/jpeg')
     })
 
     it('admin uploads valid WebP file → extension .webp derived from MIME', async () => {
@@ -208,6 +222,13 @@ describe('uploads-service', () => {
 
       expect(result).toHaveProperty('url')
       expect(result.url).toMatch(/library-games\/[0-9a-f-]+\.webp$/)
+
+      // Verify @vercel/blob put() was called with correct arguments
+      const blobPut = vi.mocked(await import('@vercel/blob')).put
+      expect(blobPut).toHaveBeenCalled()
+      const [pathname, body, options] = blobPut.mock.calls[0]
+      expect(pathname).toMatch(/^landing-media\/library-games\/[0-9a-f-]+\.webp$/)
+      expect(options?.contentType).toBe('image/webp')
     })
 
     it('admin uploads valid GIF file → extension .gif derived from MIME', async () => {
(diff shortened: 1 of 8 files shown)


tests/unit/server/uploads-service.test.ts
```diff
@@ -38,6 +38,11 @@ vi.mock('@/lib/server/shared/service-error', () => ({
   }),
 }))
 
+vi.mock('@vercel/blob', () => ({
```
Oiranca, line 41: Please use this Blob mock for the negative-path assertions too. Several validation/authorization tests still assert `mockSupabaseAdmin._uploadSpy` was not called, but that mock is no longer on the production write path after this cutover. Those tests can now pass even if `uploadLandingMediaImage()` accidentally calls `@vercel/blob.put` before rejecting. Assert `put` is not called in the non-admin, invalid-folder, invalid-MIME, oversized, empty-file, and magic-byte rejection cases, and assert the happy paths call `put` with the expected pathname/content type.

Oiranca (commented):
Storage cutover review: one testing issue to fix before merge.

tests/unit/server/uploads-service.test.ts
Oiranca: ✅ Fixed.

Replaced all vacuous Supabase storage spy assertions with actual @vercel/blob put() call/no-call assertions in uploads-service tests.

**Negative-path tests fixed (14 instances):**
- Privilege denied (non-admin member)
- Invalid folder (../etc, avatars, empty string)
- Invalid MIME (svg, pdf, html)
- File size bounds (>5MB, 0 bytes)
- Magic-byte mismatch (PNG/JPEG/WebP/GIF cross-format spoofing)

Each now asserts that put() was NOT called during validation rejection, rather than checking the now-dead Supabase spy.

**Happy-path tests enhanced (8 instances):**
- PNG, JPEG, WebP, GIF uploads (4 tests)
- 5MB boundary case
- Extension-from-MIME verification
- PNG/JPEG real-signature validation (2 tests)

Each now verifies that put() WAS called with expected pathname pattern and contentType.

**Audited for same-class defects:**
Confirmed no other test cases in the file still rely on the dead Supabase spy. Storage error handling tests correctly mock @vercel/blob.put() for error simulation.

**Validation green:**
- Suite: 1142 tests pass (includes uploads-service 32 tests)
- Typecheck: OK
- Lint: OK
- Build: OK

Commit: 4af4f1f

PR #178, opened 2026-07-25
feat(db): migrate equipment-service + database-time to Drizzle/Neon pilot (KIM-434 1/5)

Oiranca:
## KIM-434 (F3c): Swap the lib/db seam from Supabase to Drizzle/Neon — PR 1/5

Linear: https://linear.app/kimox-studio/issue/KIM-434/f3c-swap-the-libdb-seam-from-supabase-to-drizzleneon

This is PR 1 of a 5-PR stack. Base branch: `develop`.

### What this PR does
- Introduces a real Drizzle/Neon client into `lib/db/index.ts`: `getDrizzleDb()` / `getDrizzleAdminDb()`, backed by `drizzle-orm/node-postgres` + a pooled `pg.Pool` over `POSTGRES_URL` (see `.env.example`).
- Migrates the pilot service, `lib/server/equipment/equipment-service.ts`, end-to-end to idiomatic Drizzle query-builder calls.
- Migrates `lib/server/shared/database-time.ts` to support both the legacy Supabase seam and the new Drizzle seam (duck-typed on the resolved client).

### Compatibility approach (read this before reviewing)
`getDb()` / `getAdminDb()` are **unchanged** — still Supabase-backed. This is deliberate: the 11 services not yet migrated in this stack (`auth-service.ts`, `auth.ts`, `club-events-service.ts`, `events-service.ts`, `library-games-service.ts`, `saved-games-service.ts`, `partners-service.ts`, `reservations-service.ts`, `rooms-service.ts`, `tables-service.ts`, `users-service.ts`) call live Supabase query-builder chains (`.from().eq()...`, `.rpc(...)`) directly on the client these functions return. A Drizzle client cannot satisfy that API — not just at compile time, but at runtime — so swapping `getDb`/`getAdminDb` themselves would have broken those services silently, violating the "every PR in the stack must independently pass typecheck/lint/test/build" convention. Instead, new `getDrizzleDb()` / `getDrizzleAdminDb()` exports were added for genuinely migrated consumers. `database-time.ts`'s default (no-arg) call path is also kept on the legacy Supabase client, because `reservations-service.ts` (untouched, live code path in `checkUserSlotOverlap()`) calls `getDatabaseNow()` with no arguments. Once PR2-PR5 migrate the remaining 11 services, `getDb`/`getAdminDb` (and the Supabase branch in `database-time.ts`) will be removed, and `getDrizzleDb`/`getDrizzleAdminDb` renamed to `getDb`/`getAdminDb` as the final seam.

### Deliberate decision: `lib/supabase/types.ts` stays as-is for now
`lib/supabase/types.ts` (the `Database`/`Tables<T>`/`TablesInsert<T>`/`TablesUpdate<T>` types) is left completely untouched and remains the type source for the 11 not-yet-migrated services. Replacing it with Drizzle-inferred types is deferred to KIM-422 (F4, Supabase removal).

**Risk to flag:** `lib/supabase/types.ts` types are generated from the *Supabase* schema. If the Neon schema (via `lib/db/schema/*.ts`) ever diverges from the Supabase schema before KIM-422 lands, those types will silently lie for the services still using them — no compile error, wrong shape at runtime. This is an accepted, temporary risk for the duration of the KIM-434 stack.

### Manual verification (please run — not run by any agent, per DB-execution policy)
This PR does not touch database state or seed data; it only wires equipment-service.ts to point at a different database. To verify equipment records now land in Neon instead of Supabase:

1. Ensure your local `.env.local` has `POSTGRES_URL` pointed at your dev Neon/Vercel Postgres instance (see `.env.example` — do not paste the value anywhere).
2. Confirm the Drizzle schema is present in that database (this PR does not run migrations — if the `equipment`/`room_default_equipment` tables don't exist yet in Neon, apply `lib/db/migrations/*.sql` yourself first, e.g. via `drizzle-kit push` or manual `psql`, per the existing "user-only DB execution" policy).
3. Run the app locally (`pnpm dev`) and, as an admin user, create a new equipment item via the admin UI (or `POST /api/equipment` with `{"name": "Test Projector"}`).
4. In the **Neon dashboard** → your database → Tables → `equipment`: confirm the new row appears there.
5. In the **Supabase dashboard** → Table Editor → `equipment`: confirm the new row does **NOT** appear there (proves the write went to Neon, not Supabase).
6. `GET /api/equipment` (or reload the admin equipment list) and confirm the new item is returned — proves reads also go through Neon.
7. Repeat for `PATCH /api/equipment/{id}` (update), `DELETE /api/equipment/{id}` (delete), and `PUT /api/rooms/{id}/default-equipment` (room defaults) to exercise the full rewritten surface.

### Test coverage
`tests/unit/server/equipment-service.test.ts` rewritten (34 tests) to mock the Drizzle query-builder surface instead of Supabase. All existing behaviors/status codes preserved. A follow-up test-hygiene fix converted 8 error-path mocks from `mockReturnValue(Promise.reject(...))` to `mockImplementation(() => Promise.reject(...))` to eliminate unhandled-rejection noise (no behavior change).

### Validation performed
- `pnpm typecheck` — pass
- `pnpm lint` — pass
- `pnpm build` — pass (43/43 pages)
- `pnpm test` (tests/unit) — 1147 tests pass, 0 unhandled rejections

---

## Post-rebuild schema-drift check (added)

Added `scripts/check-schema-drift.mjs`, a small read-only sanity check for after the user rebuilds their Neon DB from `lib/db/migrations/0000_fine_magma.sql` + `0001_exclusion_constraints.sql` and runs `pnpm db:seed` (KIM-432). It connects via `POSTGRES_URL`, compares `information_schema.columns` against a manifest hand-derived from `lib/db/schema/*.ts`, and fails loudly with the exact table/column/type mismatch found (exits non-zero). It is plain Node ESM (no new dependency), never runs automatically, and was not executed against any real database as part of this PR.

**Run it yourself after the rebuild:**
```
pnpm check:schema-drift
```

## Auth-schema dependency check

Verified PR1's three migrated files (`lib/db/index.ts`, `lib/server/shared/database-time.ts`, `lib/server/equipment/equipment-service.ts`) for any assumption of a relationship into the `auth` schema (the old, throwaway Neon had `profiles_id_fkey`/`events_created_by_fkey`/`reservations_user_id_fkey` into `auth.users`; the rebuilt schema has no `auth` schema at all). Confirmed clean — none of the three files reference `auth.users` or join against a users/profiles table; equipment is catalog data with no user-scoping. No code changes were required.

## Split-brain disclosure

Between this PR and PR5 of the KIM-434 stack, the app talks to two databases simultaneously: Neon for whatever has been migrated so far (currently only `equipment`/`room_default_equipment` via `equipment-service.ts`), Supabase for everything else. Three cross-table joins span that boundary and are inconsistent by construction during this window: `event_equipment` (equipment↔events), `room_default_equipment` (equipment↔rooms, partially migrated — the `equipment` side is on Neon but `rooms` is still on Supabase until PR2), `reservation_equipment` (equipment↔reservations). This PR's changes were verified in isolation (equipment CRUD + room-default-equipment assignment against Neon); this PR does **not** claim the app works end-to-end — the app remains split-brain until PR5 lands and all 13 services are on the same database.

### DIFF
--- a/lib/server/equipment/equipment-service.ts
+++ b/lib/server/equipment/equipment-service.ts
@@ -1,7 +1,7 @@
 import { asc, eq, inArray } from 'drizzle-orm'
 import { getDrizzleAdminDb, getDrizzleDb } from '@/lib/db'
 import { equipment, roomDefaultEquipment } from '@/lib/db/schema'
-import { serviceError } from '@/lib/server/shared/service-error'
+import { ServiceError, serviceError } from '@/lib/server/shared/service-error'
 import { ERROR_CODES } from '@/lib/types/error-codes'
 import type { Equipment } from '@/lib/types'
 import type { SessionUser } from '@/lib/server/auth/auth'
@@ -44,6 +44,29 @@ async function runQuery<T>(query: Promise<T>): Promise<T> {
   } catch {
     serviceError('Internal server error', 500)
   }
+}
+
+/**
+ * Name of the DB-level unique constraint added on
+ * `room_default_equipment.equipment_id` (KIM-434 PR1 hotfix, see
+ * lib/db/migrations/0002_room_default_equipment_equipment_unique.sql and the
+ * schema comment in lib/db/schema/equipment.ts). Used to recognize a
+ * Postgres 23505 (unique_violation) raised specifically by this constraint,
+ * so it can be translated into the existing EQUIPMENT_LOCKED_TO_ANOTHER_ROOM
+ * business error instead of surfacing as an unhandled 500.
+ */
+const ROOM_DEFAULT_EQUIPMENT_UNIQUE_CONSTRAINT = 'room_default_equipment_equipment_id_unique'
+
+/**
+ * `node-postgres` (the driver behind the Drizzle/Neon seam) throws raw
+ * `pg` `DatabaseError` instances with a `code`/`constraint` shape on
+ * constraint violations; it does not wrap them. Duck-typed here rather than
+ * importing `pg`'s error class to keep this check dependency-light.
+ */
+function isEquipmentExclusivityViolation(error: unknown): boolean {
+  if (typeof error !== 'object' || error === null) return false
+  const { code, constraint } = error as { code?: unknown; constraint?: unknown }
+  return code === '23505' && constraint === ROOM_DEFAULT_EQUIPMENT_UNIQUE_CONSTRAINT
 }
 
 function toEquipment(row: EquipmentRow): Equipment {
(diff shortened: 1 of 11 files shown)


lib/server/equipment/equipment-service.ts
```diff
@@ -156,51 +160,37 @@ export async function setRoomDefaultEquipment(
...
+    )
+
+    const conflicts = existingDefaults.filter((row) => row.roomId !== roomId)
 
     if (conflicts.length > 0) {
       serviceError(ERROR_CODES.EQUIPMENT_LOCKED_TO_ANOTHER_ROOM, 400)
     }
   }
 
   // Delete existing defaults for this room
```
Oiranca: Please wrap the conflict check, delete, and insert in a single Drizzle transaction. As written, `setRoomDefaultEquipment()` deletes the room's existing defaults and then inserts replacements in separate statements; if the insert fails because one id is invalid, a FK is missing, or the connection drops, the service returns 500 after already clearing the room defaults. This became easier to fix with Drizzle, and the test suite should cover insert-failure rollback so we do not regress this behavior during the remaining DB migration PRs.

Oiranca (commented):
Drizzle pilot review: one data-consistency issue to fix before merge.

lib/server/equipment/equipment-service.ts
Oiranca: ✅ Fixed in d4c72de. setRoomDefaultEquipment() now wraps the conflict-check SELECT, DELETE, and INSERT in a single db.transaction() to close the data-loss window — if the INSERT fails, the DELETE rolls back too instead of leaving the room's default-equipment config silently cleared. For the concurrent-race exclusivity guarantee, added a DB-level unique constraint on room_default_equipment.equipment_id (migration 0002_room_default_equipment_equipment_unique.sql, prepared but not applied — user applies manually). A transaction alone doesn't prevent two concurrent transactions from both passing the SELECT check under READ COMMITTED isolation, so the unique constraint is the real guarantee; the transaction ensures the losing request fails cleanly (400 EQUIPMENT_LOCKED_TO_ANOTHER_ROOM, via a new isEquipmentExclusivityViolation() translation of the Postgres 23505 unique-violation) instead of a raw 500, and ensures no partial DELETE-without-INSERT state on any failure. Added regression tests covering the rollback-on-failure path, the unique-violation-to-400 translation, and a negative case proving the constraint-name match doesn't over-match unrelated unique violations. Independently re-verified by security-reviewer: transaction wraps all three ops correctly, requireAdminSession() unchanged, constraint name consistent across schema/migration/service code, full CI green (37/37 equipment tests, 1150/1150 suite, typecheck/lint/build).

PR #182, opened 2026-07-26
feat(db): migrate events-service + library-games-service to Drizzle/Neon (KIM-434 3/5)

Oiranca:
## Summary

- Migrates `lib/server/events/events-service.ts` and `lib/server/games/library-games-service.ts` from the legacy Supabase seam to the Drizzle/Neon seam (`getDrizzleDb()` / `getDrizzleAdminDb()`), following the pattern established in PR1 (`equipment-service.ts`) and PR2 (`tables-service.ts`).
- All admin-privilege checks (`role === 'admin'`) remain enforced in the service layer via `requireAdminSession()`, unchanged in intent from the pre-migration Supabase logic.
- `reservations` is not yet migrated (deferred to KIM-434 PR5), so event/library-games writes that need to cancel/count overlapping reservations still call the legacy Supabase admin client for that one table — see the "Split-brain disclosure" doc comment on `cancelOverlappingReservationsForBlocks()` in `events-service.ts` for the full consistency tradeoff this introduces during the migration window.
- Adds/extends `tests/unit/mocks/drizzle-mock.ts`: a new transaction-aware Drizzle mock helper (`createTransactionAwareMockBuilder`) with per-table insert dispatch and fixture-state rollback verification, used by the migrated test suites to correctly simulate `db.transaction()` semantics.
- `lib/server/events/club-events-service.ts` is intentionally **NOT migrated** in this PR — it contains only a 22-line compatibility shim (3 call-site adaptations for the new camelCase `isClubEventRow({ titleEs, titleEn })` signature and the new `deleteEventCascade(id)` signature) so it keeps compiling against the migrated `events-service.ts`. Its full migration is deferred to **KIM-438**, which is blocked on this PR merging.

## Scope

- `lib/server/events/events-service.ts` — full Drizzle migration
- `lib/server/games/library-games-service.ts` — full Drizzle migration
- `lib/server/events/club-events-service.ts` — compatibility shim only (not a migration)
- `tests/unit/mocks/drizzle-mock.ts` — new transaction-aware mock helper
- Test files updated for the new seam: `club-events-service.test.ts`, `equipment-service.test.ts`, `events-preview.test.ts`, `events-service-multiday.test.ts`, `events-service.test.ts`, `library-games-service.test.ts`, `oir208-unified-events.test.ts`

## Validation (re-run independently by security-reviewer in-worktree)

- `pnpm typecheck` — clean
- `pnpm lint` — `✔ No ESLint warnings or errors`
- `pnpm test` — `Test Files 76 passed (76)` / `Tests 1162 passed (1162)`
- `pnpm build` — exit code 0

## Security review

- No secrets or credentials found in the diff (including `.claude/agent-progress.md` additions).
- Admin/session/ownership checks preserved in the service layer for every mutation in both migrated files.
- No raw/string-concatenated SQL — all queries use the Drizzle query builder (parameterized).
- No test file makes a real DB call — `@/lib/db` is mocked in every touched test file; no live connection, container, or embedded Postgres.
- `events-service.ts` / `library-games-service.ts` do not perform member-scoped reads of `reservations`/`saved_games` on behalf of a session user (the only `reservations` access is admin-driven overlap cancellation/counting during event scheduling, not a per-member reservation list), so `assertMemberRowsScoped()` does not apply here.

Refs: KIM-434 (3/5). Follow-up: KIM-438 (club-events-service.ts full migration, blocked on this PR).

## Update: saved_games cancellation cascade (review follow-up)

An internal security-reviewer audit of dropped Supabase triggers/functions (not part of the original Codex review pass) found a second regression beyond the room/table guard fixed in the prior round: the `event_blocks_cancel_saved_games` trigger (`AFTER INSERT OR UPDATE OF room_id, date ON event_room_blocks`) had no Drizzle/Neon equivalent — creating or updating an admin event with room blocks had silently stopped cancelling active `saved_games` rows for the blocked room/date.

**Fix:** new exported `cancelSavedGamesForBlockedRoom(tx, blocks)` in `lib/server/events/events-service.ts`, room-scoped (not table-scoped, matching the trigger's `NEW.room_id` semantics), wired into all 4 `event_room_blocks`-insert transactions (`createEventAtomic`, `updateEventAtomic`, `createEventWithBlocksAtomic`, `updateEventWithBlocksAtomic`). Runs inside the same Drizzle transaction as the block insert, giving true all-or-nothing atomicity (a rollback of the block write also rolls back the cascade) — unlike the still-Supabase `reservations` cancellation, which must run post-commit.

**Advisory-lock decision** (quoted verbatim from the code comment directly above `cancelSavedGamesForBlockedRoom` in `events-service.ts`):

> Concurrency guard (reimplemented, not dropped — see decision writeup in the PR description / handoff notes for the full reasoning): the `pg_advisory_xact_lock` loop IS reimplemented below via `tx.execute(sql\`...\`)`, in the same per-table, id-ascending order the trigger used. Investigation (reading `saved_games_no_active_overlap`'s actual definition in `supabase/migrations/20260619000003_kim384_create_saved_games.sql`) showed it is NOT a substitute for this lock: that EXCLUDE constraint only prevents two `saved_games` rows from overlapping on the same table_id + daterange while both `active` — it says nothing about a `saved_games` row overlapping an `event_room_blocks` row. The actual race this lock guards against is between THIS cascade and `validate_saved_game`'s BEFORE INSERT/UPDATE trigger on `saved_games` (`supabase/migrations/20260619000006_kim384_saved_game_validation_trigger.sql`), which takes the SAME advisory lock key (`hashtextextended(table_id::text, 0)`) before checking for an event-block conflict (`SAVED_GAME_EVENT_CONFLICT`). Without a matching lock here, a member's "create saved_game" transaction and an admin's "create/update event block that cancels saved_games" transaction can each run their conflict check against data that doesn't yet reflect the other's uncommitted write (both under READ COMMITTED), letting a new active saved_games row survive uncancelled for a room/date that's actually blocked, or vice versa. Taking the same advisory lock forces one of the two transactions to block until the other commits, so whichever runs second always sees the final, committed state of the other. This IS reimplemented for that reason.

**Validation for this round:**
- `pnpm typecheck` — clean
- `pnpm lint` — `✔ No ESLint warnings or errors`
- `pnpm test` — `Test Files 76 passed (76)` / `Tests 1169 passed (1169)` (baseline 1166 + 3 new regression tests)
- `pnpm build` — exit code 0, `✓ Compiled successfully`

**Security review of this round:** the `tx.execute(sql\`SELECT pg_advisory_xact_lock(hashtextextended(${tableId}::text, 0))\`)` call was scrutinized as the only raw-SQL escape hatch in this PR — confirmed it uses Drizzle's `sql` tagged template, which parameterizes `${tableId}` as a bound placeholder rather than string-concatenating it into raw SQL text (verified directly in the code, and independently confirmed by a new regression test asserting the interpolated value lands in a separate `queryChunks` slot, not inlined into the SQL string). No secrets/credential leaks found in the diff. New tests use mocks only (`tests/unit/mocks/drizzle-mock.ts`'s new `executeMock`/`tx.execute` support) — no real DB connection. This cascade runs inside the existing admin-gated transactions and does not introduce a new entry point or bypass `requireAdminSession()`.

Reuse note: `cancelSavedGamesForBlockedRoom` is exported so **KIM-438** (`club-events-service.ts`, tracked separately, not in this PR) can call it from its own room-block write path once wired, instead of duplicating this logic.


lib/server/events/events-service.ts
```diff
@@ -260,6 +217,393 @@ export function validateAndNormaliseSchedule(
...
+
+      const blocksWithRoom = normBlocks.filter(
+        (b): b is NormalisedEventSchedule & { room_id: string } => b.room_id !== null,
+      )
+
+      let insertedBlocks: EventRoomBlockRow[] = []
+      if (blocksWithRoom.length > 0) {
+        insertedBlocks = await tx
+          .insert(eventRoomBlocks)
+          .values(
```
Oiranca, line 629: This Drizzle replacement for the RPC writes `roomId` and `tableId` directly, but it does not reimplement the RPC's room/table consistency guard. The existing migration notes say independent FKs are not enough: a caller can pair `table_id` from room A with `room_id` from room B unless the service checks `tables.id = tableId AND tables.room_id = roomId` before insert. That matters for availability and cancellation logic, because the row then says one room while pointing at a table in another.

Please validate all non-null `table_id` values against their submitted `room_id` inside the same transaction before inserting blocks, and return a 400 for mismatches. Add the same coverage for both create and update multi-block paths.

lib/server/events/events-service.ts
```diff
@@ -649,25 +872,26 @@ export async function previewEventConflicts(body: {
     return { total: 0, blocks: [] }
   }
 
-  const admin = getAdminDb()
+  const drizzleDb = getDrizzleDb()
 
   // Pre-fetch table ids for all distinct rooms in one round-trip (avoids N+1)
```
Oiranca, line 1064: The conflict preview prefetches all tables for each room, but the later count path still uses the room's full table list for every block. That ignores `block.table_id`, while `cancelOverlappingReservationsForBlocks()` correctly scopes a table-level block to `[block.tableId]`. For table-scoped schedules the preview will over-report conflicts for sibling tables, so admins may cancel or avoid changes based on false conflict counts.

Please use `block.table_id ? [block.table_id] : roomTableMap.get(block.room_id) ?? []` when building the reservations count query, and add a regression test proving a table-scoped schedule only sends that one table id into `.in('table_id', ...)`.

Oiranca (commented):
Reviewed for best practices, security, and test coverage. I found correctness/security gaps around table-scoped event blocks that should be fixed before merge.

lib/server/events/events-service.ts
Oiranca: Fixed in commit a4c648a — added `assertBlocksTableRoomConsistency()` in `lib/server/events/events-service.ts`, called inside the same Drizzle transaction as the insert/update in both `createEventWithBlocksAtomic` and `updateEventWithBlocksAtomic`. It queries the actual `tables` table for every block's declared `table_id`, compares the returned `room_id` against the block's declared `room_id`, and throws a 400 `ServiceError` on any mismatch before the write proceeds — reimplementing the guard the old `create_event_with_blocks`/`update_event_with_blocks` RPCs used to enforce. Regression tests added in commit d6b1ddd (`tests/unit/server/events-service-multiday.test.ts`) cover reject-on-create, reject-on-update, and the allow-on-match path. ✅ Resolved.

lib/server/events/events-service.ts
Oiranca: Fixed in commit a4c648a — `previewEventConflicts()` in `lib/server/events/events-service.ts` now scopes the conflict count to `[block.table_id]` when a block declares a specific table, falling back to the room's full table list only when `table_id` is null (whole-room block). This matches the scoping actually used by `cancelOverlappingReservationsForBlocks()`, so the preview count no longer overstates reservations that will be cancelled for a table-level block. Regression test added in commit d6b1ddd (`tests/unit/server/events-preview.test.ts`) asserts the count query is filtered to the single table, not the whole room. ✅ Resolved.

Oiranca:
## Security audit follow-up: saved_games cancellation cascade (beyond original review scope)

This comment covers an addition that goes beyond what the original Codex review pass flagged in this PR. It was found via a proactive internal audit of Supabase triggers/functions dropped by this migration, not by an inline review comment on the existing diff — so this is a standalone note rather than a reply to any existing thread.

**Finding:** the `event_blocks_cancel_saved_games` trigger had no Drizzle/Neon equivalent after this migration — creating/updating an admin event with room blocks had silently stopped cancelling active `saved_games` rows for the blocked room/date.

**Fix (3 new commits on top of this PR's prior tip):**
- `78ae3a7` — new `cancelSavedGamesForBlockedRoom(tx, blocks)` reimplementing the dropped cascade, room-scoped, wired into all 4 event_room_blocks-insert transactions.
- `c083fbf` — reimplements the trigger's `pg_advisory_xact_lock` loop to close a real concurrency race against the Supabase-side `validate_saved_game` trigger, which takes the same lock key.
- `2aec355` — adds `tx.execute` mock support to `tests/unit/mocks/drizzle-mock.ts` and 3 regression tests covering room-scoped cancellation, no-op with no overlap, and no-op with no tables.

Both the cascade and the advisory lock are now reimplemented with regression tests. Independently re-verified: `pnpm typecheck` clean, `pnpm lint` clean, `pnpm test` — `Tests 1169 passed (1169)`, `pnpm build` exit 0. Full reasoning and the verbatim advisory-lock decision comment are in the updated PR description above.

lib/server/events/events-service.ts
```diff
@@ -260,6 +217,580 @@ export function validateAndNormaliseSchedule(
...
+//
+// Test-infra gap (does not block the fix above): tests/unit/mocks/drizzle-mock.ts's
+// shared transaction builder has no chain for `tx.execute(...)`, so
+// `pnpm test` is expected to newly fail wherever a test exercises this
+// function against the mock. Per repo convention test infra is
+// qa-engineer-owned, not software-engineer's — qa-engineer should add a
+// `tx.execute` mock (e.g. resolving to `{ rows: [] }`) to close that gap.
+//
+// Exported (same pattern as deleteEventCascade below) so
+// lib/server/events/club-events-service.ts can reuse it once its own
```
Oiranca, line 384: Deferring this to KIM-438 leaves the same saved_games regression open for the current club-events room-block path. `club-events-service.ts` still creates/replaces room blocks through `apply_club_event_room_blocks`; after `saved_games` moved to the Drizzle/Neon seam, that legacy Supabase RPC/trigger cannot cancel the Neon `saved_games` rows. So internal events now get the new cascade, but club events with `blocksRooms` can still leave active saved games in a blocked room/date until a later migration. Please either wire this cascade into the current club-event block replacement path in this PR, or explicitly block/disable club-event room blocking until KIM-438 lands. Add regression coverage around `createClubEvent`/`updateClubEvent` with `blocksRooms: true` proving active saved_games for the blocked room/date are cancelled.

Oiranca (commented):
Follow-up review after the saved_games cascade update: one remaining coverage gap affects current club-event room-block writes.

lib/server/events/events-service.ts
Oiranca: You've identified a real hazard, and I'm not dismissing it. The disagreement is narrow: not *whether* the club-event path needs this cascade, but *which PR is mechanically capable of adding it* and *when the exposure actually becomes live*. I verified each point below rather than reasoning from the diff.

## 1. There is no live regression today — both sides are still on the Supabase seam

The premise of the finding is "after `saved_games` moved to the Drizzle/Neon seam." That move hasn't happened yet on `develop`.

- `club-events-service.ts` still writes blocks through the legacy RPC, unchanged by this PR — line 499:
  ```ts
  const { data, error } = await admin.rpc('apply_club_event_room_blocks', {
  ```
  `git diff develop origin/migration-f3c-03-content-batch -- lib/server/events/club-events-service.ts` returns only three compatibility shims (two `isClubEventRow` camelCase mappings, one `deleteEventCascade(id)` signature adaptation). None touch the block-write path.

- `saved_games` is still on Supabase. `saved-games-service.ts` on `develop` uses `getAdminDb()` at 5 call sites (lines 82, 122, 156, 173, 217), and `lib/db/index.ts:123-124` resolves that to Supabase:
  ```ts
  export function getAdminDb(): LegacySupabaseDbClient {
    return createSupabaseServerAdminClient()
  }
  ```

So the club-event path writes `event_room_blocks` into Supabase, the `event_blocks_cancel_saved_games` trigger fires there, and it cancels `saved_games` rows in that same database. That path is intact.

The asymmetry you spotted is real and worth stating precisely: the internal-events path regressed **because** this PR moved its `event_room_blocks` writes to Neon, so the Supabase trigger stopped firing for it. The club-event path still writes through the RPC into Supabase, so its trigger still fires. That is exactly why one needed the cascade reimplemented and the other doesn't yet.

## 2. The gap opens at a specific PR, not "later"

It opens the moment `saved_games` moves to Neon — that is **KIM-441**, branch `migration-f3c-06-saved-games`, commit `91713fd`, which swaps the service to `getDrizzleAdminDb()`. I verified it is **not** merged into `develop` (`git merge-base --is-ancestor` returns false) and it currently has no open PR. Until it lands, the RPC and the rows it must cancel are in the same database.

## 3. Option 1 is not mechanically available in this PR

`cancelSavedGamesForBlockedRoom(tx: AdminTx, blocks)` takes a Drizzle transaction handle:

```ts
type AdminTx = Parameters<Parameters<AdminDbClient['transaction']>[0]>[0]
// AdminDbClient = NodePgDatabase<typeof schema>
```

The club-event block replacement executes inside a Supabase `SECURITY DEFINER` plpgsql function, in a different database. There is no way to pass a Drizzle transaction handle into it.

Satisfying option 1 today would therefore require a **second, Supabase-side implementation** of the same cancellation semantics — the room-scoped candidate selection, the `pg_advisory_xact_lock` ordering, and the status transition — living in parallel with the Drizzle one. Two divergent copies of a single cancellation rule, guaranteed to drift, is precisely the duplication failure mode this migration has spent its entire course eliminating. That is the main reason not to do it, independent of scheduling.

## 4. Option 2 costs more than the ordering constraint saves

Disabling club-event room blocking would remove working, user-facing functionality to guard a window that is not currently open and that merge ordering closes at zero cost.

The alternative I'm committing to instead: **KIM-438 merges before or together with KIM-441.** I am recording this as a blocking relation in Linear so it survives past this thread rather than depending on anyone remembering it. Merging KIM-441 on its own is the only sequence that opens the window, and that relation is what prevents it.

## 5. The regression coverage you asked for lands in KIM-438

Tests around `createClubEvent` / `updateClubEvent` with `blocksRooms: true`, asserting that active `saved_games` for the blocked room and date are cancelled, will be in KIM-438's PR — because that is the PR where the behaviour will exist to be tested. Asserting it here would mean asserting against the Supabase trigger, which is not this PR's code and is about to be replaced.

This is scheduled work with an identified owner, not an intention. KIM-438 has its wiring point located — `applyClubEventRoomBlocksAndMaterialsTx()` at line 500, immediately after the `.insert(eventRoomBlocks)` at line 540 — and type compatibility is confirmed: that function already takes `DbTransaction` from `@/lib/db`, and both it and `AdminTx` derive from `NodePgDatabase<typeof schema>`, so `cancelSavedGamesForBlockedRoom(tx, blocks)` can be called there unchanged. The helper was deliberately exported rather than kept private for exactly this reuse.

---

**Status: confirmed, fix scheduled in KIM-438, exposure closed by merge ordering.** Not marking this resolved, since nothing was changed in this PR in response to it.

Oiranca: Two concrete follow-ups to the above, so nothing here rests on my word alone.

**The blocking relation is recorded, not just promised.** KIM-441 now carries `blockedBy: KIM-438` in Linear (verified on the issue itself, not asserted). Merging KIM-441 on its own — the only sequence that opens the window described above — is now gated on KIM-438 landing first.

**The acceptance criteria for KIM-438's regression coverage, stated concretely**, since "tests will be added" is not a verifiable commitment:

> Tests under both `createClubEvent` and `updateClubEvent` with `blocksRooms: true`, each seeding an **active** `saved_games` row on a table in the blocked room with a date range overlapping the block's date, and asserting that row's status becomes `cancelled` after the call.

That shape is deliberate, and it's the part that answers your actual concern: it proves the helper **fires** on the club-event path against real seeded state — not merely that it compiles, that it's imported, or that the call site exists. A silently unwired helper, a call placed outside the transaction, or a wrong argument shape (for instance passing table-scoped rather than room-scoped blocks) all still produce a green build and a passing import; they do not survive an assertion that a specific seeded row flipped to `cancelled`.

Worth noting explicitly, since it is the likeliest way this could go wrong quietly: the cancellation is **room-scoped, not table-scoped** — every table in the blocked room is in range, not only the table named on the block. The seeded row in those tests sits on a *different* table in the same room for exactly that reason.

PR #183, opened 2026-07-26
feat(db): migrate users-service to Drizzle/Neon (KIM-440)

Oiranca:
## Summary

- Migrates `lib/server/users/users-service.ts` off Supabase `getAdminDb()` to `getDrizzleAdminDb()` + drizzle-orm for all `profiles` table CRUD (list/search/update/import/resetNoShows/unblockUser/deleteUser). Supabase Auth admin calls (`createAuthUser`/`deleteAuthUser`/`updateAuthUserById`) remain on the Supabase seam via a dedicated `createSupabaseServerAdminClient()` — that's a separate concern (Auth, not Postgres/`profiles`) and out of scope for this migration.
- Migrates `lib/server/users/profile-mappers.ts`: `PublicProfileRow` now derives from Drizzle's `typeof profiles.$inferSelect` (camelCase) instead of Supabase's `Tables<'profiles'>` (snake_case). `toPublicUser()` now converts `Date` → ISO string for timestamp fields, with null-safe handling for `activeFrom`/`blockedUntil`.
- **Deliberate side-effect, not scope creep:** `lib/server/auth/auth-service.ts` previously imported the shared `toPublicUser`/`PublicProfileRow` from `profile-mappers.ts`. Since that file stays on the legacy Supabase seam (snake_case rows) for now, changing the shared mapper's shape would have silently type-checked but broken at runtime there. Decoupled `auth-service.ts` with its own local snake_case `PublicProfileRow` alias and `toUser()` mapper — field-for-field identical to the old shared mapper, zero behavior change to `auth-service.ts`'s public surface.
- Test infrastructure: extended the shared Drizzle query-builder mock (`tests/unit/mocks/drizzle-mock.ts`) to support `.limit()/.offset()` pagination chains and update-without-`.returning()` (fire-and-forget) calls; updated `users-service.test.ts`, `profile-mappers.test.ts`, and `member-import.test.ts` mocks accordingly. Mocks only — no real DB/Postgres connection in any test.

## Security review (self-performed, security-reviewer agent)

- **Privilege checks preserved:** all 7 exported functions in `users-service.ts` (`importMembersFromCsv`, `importMembersFromSource`, `listPaginatedUsers`, `updateUser`, `resetNoShows`, `unblockUser`, `deleteUser`) call `requireAdminSession(session)` at the top — gating stays in the service layer, not route handlers.
- **`assertMemberRowsScoped()` exemption verified:** read every function in the file — all are admin-gated, cross-member CRUD with no member-scoped "fetch my own rows" read path, so the defense-in-depth scoping helper correctly does not apply here (member self-service reads live in `lib/server/auth/*`, untouched).
- **`auth-service.ts` decoupling verified clean:** new local `toUser()` maps the identical field set as the old shared `toPublicUser()` — no change to what fields are exposed to callers.
- **No secrets/credentials** found in the diff (scanned for hardcoded passwords, API keys, connection strings — none present; temp password generation uses `crypto.randomUUID()`, no logging of the value).
- **SQL injection safety:** all queries go through drizzle-orm's parameterized query builder (`eq`, `ilike`, `or`) — no raw string interpolation. Search term sanitization (`sanitizeSearchTerm`/`escapeLikeWildcards`) preserved from the pre-migration version.
- **Error handling parity:** auth-user rollback on profile-persist failure (member import) and profile rollback on auth-email-update failure (updateUser) both preserved, matching prior Supabase-era behavior including the "best-effort, proceed to 500 regardless" semantics on rollback failure.
- **Tests never touch a real DB:** confirmed the extended `drizzle-mock.ts` is a pure in-memory chainable mock (vi.fn-based), no live connection, no test DB, no embedded Postgres.

## Test plan

Re-ran independently in an isolated `git worktree` (not the shared checkout):

- `pnpm test` → **1170/1170 passed** (76 test files)
- `pnpm typecheck` → clean, 0 errors
- `pnpm lint` → clean, 0 warnings/errors
- `pnpm build` → exit 0, 43/43 routes generated

Test counts vs `develop` unchanged: `users-service.test.ts` 37/37, `profile-mappers.test.ts` 8/8, `member-import.test.ts` 24/24.

Closes KIM-440.

## Codex review follow-up (KIM-440)

Reviewer `Oiranca` (Codex) flagged three findings, all addressed:

- **Real production bug found and fixed:** new-member import persisted nothing on Neon. The `profiles` row used to be created by the Supabase DB trigger `on_auth_user_created` → `handle_new_user()` (`supabase/migrations/20260417000003_baseline.sql:216-232`), which lives in the Supabase database — not Neon, where `getDrizzleAdminDb()` connects. The service's `UPDATE profiles ... WHERE id = authData.user.id` therefore matched zero rows, silently failing import and rolling back the auth user. Fixed with an explicit `db.insert(profiles).values(...).returning(...)`.
- **Test mocks corrected to model real DB semantics:** the mock previously created a profile row on any `UPDATE`, regardless of whether a matching row existed — masking the bug above. The mock now only creates rows on explicit INSERT, matching real Postgres `UPDATE ... WHERE` semantics. The prior diagnostic test proving the failure was inverted into a permanent regression guard.
- **Test precision gap fixed (no underlying production bug):** `users-service.test.ts`'s pagination test shared one `selectMock` response between the data query and the `count()` query, so a mismatch between returned rows and reported total could not have been caught. Fixed to return distinct responses per query; service behavior was already correct here.
- **Tripwire flagged for follow-up (not fixed in this PR):** `register()` in `lib/server/auth/auth-service.ts` has the identical `createAuthUser`-then-UPDATE pattern and still depends on the same Supabase trigger. It isn't broken today because it still runs against the legacy Supabase client, but it will reproduce this exact bug once `auth-service.ts` migrates to the Drizzle/Neon seam. Tracked under KIM-444.

Final validation after all fixes: `pnpm test` → **1172/1172 passed** (76 test files), `pnpm typecheck` clean, `pnpm lint` clean, `pnpm build` exit 0 (43 routes). Test counts vs `develop`: `member-import.test.ts` 24→25, `users-service.test.ts` 37→38 — none removed.


lib/server/users/users-service.ts
```diff
@@ -192,24 +187,29 @@ async function importMembersFromNormalizedRows(input: {
...
-      })
-      .eq('id', authData.user.id)
-      .select(PROFILE_COLUMNS)
-      .maybeSingle()
-
-    if (updateProfileError || !persistedProfile) {
-      await deleteAuthUser(admin, authData.user.id)
+    let persistedProfile: { id: string } | undefined
+    try {
+      ;[persistedProfile] = await db
```
Oiranca, line 198: This new-member import path still assumes that `createAuthUser()` creates a matching `profiles` row in the database used by `getDrizzleAdminDb()`, then updates that row by `authData.user.id`. That was true for the legacy Supabase DB trigger, but `getDrizzleAdminDb()` is documented as the Neon/Postgres seam, where the Supabase Auth trigger does not create this row. In the migrated path this will return no rows, delete the auth user, and mark every new member import as `persist_import_failed`.

Please insert the profile row into Neon explicitly after the auth user is created (using the auth user id as `profiles.id`), and keep the compensating `deleteAuthUser()` rollback if the insert fails. Add a regression test where `update(...).returning()` returns `[]` for the create path so the current false-positive mock cannot hide this.

tests/unit/server/users-service.test.ts
```diff
@@ -1,715 +1,507 @@
...
     vi.resetModules()
     vi.clearAllMocks()
-    resetQueryMocks()
+    resetMocks()
   })
 
-  it('clamps page=0 to 1', async () => {
+  it("clamps page=0 to 1", async () => {
     const adminSession = createAdminSession()
+    selectMock.mockResolvedValue(profileRows.slice(0, 2))
```
Oiranca: These list tests use the same `selectMock` response for both the data query and the `count()` query, so `countRows[0]?.value` is `undefined` and the service can return `data.length > 0` with `total === 0` without any test failing. Please make the Drizzle mock distinguish row selects from count selects, and assert `total`/`totalPages` for non-empty and searched results. That will catch pagination regressions instead of only proving the chain was callable.

Oiranca (commented):
Reviewed for best practices, security, and test coverage. I found a blocking correctness issue plus test gaps that should be addressed before merging.

tests/unit/server/member-import.test.ts
```diff
@@ -52,41 +68,94 @@ function resetProfileState() {
...
+ *  - "create new member" payloads always include `memberNumber` (see
+ *    `importMembersFromNormalizedRows`'s create-path `.set({ memberNumber, ... })`)
+ *    -> build a fresh row keyed by that member number.
+ *  - "update existing member" payloads never include `memberNumber` (only
+ *    `fullName`/`email`/`phone`) -> merge into the single pre-seeded fixture.
+ * No test in these describes mixes both scenarios in one call, so keying
+ * off `updates.memberNumber` presence is sufficient without needing to
+ * thread the real row id through the mock's `.where()` clause.
+ */
+function configureImportUpdateMock() {
```
Oiranca: This mock models the create path by creating a profile row when `updateMock` sees `updates.memberNumber`, which reproduces the service's assumption instead of validating real Drizzle behavior. A real `UPDATE profiles WHERE id = authUser.id` cannot create a row in Neon. Please change the test harness so create-vs-update semantics are explicit: new imports should exercise `insert(profiles).values(...).returning(...)`, while existing imports should exercise `update(...).where(...)`. Add a failing case for `returning([])` after auth creation and assert that `deleteAuthUser()` is called.

lib/server/users/users-service.ts
Oiranca: ✅ Confirmed as a real production bug, not a false positive — you were right.

Root cause pinned: the `profiles` row was previously created by a Supabase DB trigger, `on_auth_user_created` → `handle_new_user()`, defined at `supabase/migrations/20260417000003_baseline.sql:216-232`, which did `INSERT INTO public.profiles (id, email, auth_email, member_number, role)`. That trigger lives in the Supabase database; `getDrizzleAdminDb()` connects to Neon via `POSTGRES_URL` — a different database. So `UPDATE profiles ... WHERE id = authData.user.id` matched zero rows and new-member import persisted nothing, falling into the service's own `persist_import_failed` branch and rolling back the auth user.

Fix in `b41f03b`: replaced that UPDATE with an explicit `db.insert(profiles).values({...}).returning({ id: profiles.id })`. Column set verified against `lib/db/schema/profiles.ts` (`id`, `memberNumber`, `authEmail` are NOT NULL and all supplied) and against what the vanished trigger used to insert, plus the fields the old UPDATE already set (`fullName`, `phone`, `isActive: false`, `activeFrom: null`, `pswChanged: null`). Follows the existing `.insert(...).values(...).returning()` convention already used in `equipment-service.ts` and `partners-service.ts`.

No `onConflictDoUpdate`: re-import of an existing member number is already routed to the separate UPDATE branch by the prior `SELECT ... WHERE memberNumber = ...` existence check, so an upsert isn't needed for that path.

Fail-first evidence: with the corrected mocks but before the fix, 4 tests failed with `expected +0 to be 1` (and `to be 60` for the bulk case). After the fix plus the mock-observability fix, all pass.

Related tripwire found and reported separately (NOT changed in this PR): `register()` in `lib/server/auth/auth-service.ts` (~lines 618-693) has the identical `createAuthUser`-then-UPDATE pattern and an explicit comment saying it relies on `on_auth_user_created`. It is not currently broken because it still goes through `getAdminDb()` (legacy Supabase client, where the trigger still exists), but it will reproduce this bug verbatim the moment auth-service migrates to the Drizzle/Neon seam. Tracked for the auth-service migration (KIM-444 scope). `activateAccount()` does not call `createAuthUser` and has no such dependency.

tests/unit/server/member-import.test.ts
Oiranca: ✅ Agreed completely, and this was the most important of the three: the mock encoded the service's assumption rather than real database behavior, which is exactly why a 1170/1170 green suite sat on top of broken production code.

Fixed in `a60c28d`: the harness no longer creates a profile row when `updateMock` sees `updates.memberNumber`. An `UPDATE ... WHERE id = ...` that matches no existing row now affects zero rows and creates nothing, matching real Postgres semantics. Only an explicit INSERT creates a row.

That correction is what made the bug visible: against the unfixed service it produced 4 genuine failures, all `expected +0 to be N`, plus a dedicated isolating test asserting `persist_import_failed`, `deleteAuthUser` called with the new auth id, and `profileState.get('100020')` undefined — which ruled out mock miswiring as the cause.

Follow-on in `46c440e`: closing the loop required the shared helper to actually observe inserts. `tests/unit/mocks/drizzle-mock.ts` was threading `.values(...)` nowhere, so a correct insert resolved to bare `insertMock()` → `undefined`. That payload is now forwarded the same way the update chain forwards `updates`/where-args, and a `configureImportInsertMock()` was added mirroring the existing `configureImportUpdateMock()`.

The temporary diagnostic test that asserted the broken behavior has been inverted into a permanent regression guard, renamed "persists a new member via explicit INSERT (KIM-440 Finding A regression guard)", asserting `createdCount === 1`, no `persist_import_failed` issue, `deleteAuthUser` NOT called, and the new row present in the fixture.

Test count went 24 → 25; nothing removed.

tests/unit/server/users-service.test.ts
Oiranca: ✅ Fixed: the data query and the `count()` query no longer share one `selectMock` response; they now get distinct responses per query, so a mismatch between returned rows and reported total is actually caught. Test count went 37 → 38.

Important distinction, stated plainly: unlike finding #1, this one did NOT correspond to a live production bug. After correcting the mock, `users-service.test.ts` runs 38/38 green against the current service — the service's pagination/count behavior was already correct. This was a genuine test-precision gap: the prior assertion could not have caught a regression of that shape, so the fix is real and worth having, but we are not claiming it repaired broken behavior.

An audit for the same "mock encodes the assumption" pattern was run across the sibling test files. `profile-mappers.test.ts` has no DB mock usage (pure mapper functions). `users-service.test.ts` uses only per-test static stubs scoped to individual scenarios, with no payload-inferring `mockImplementation` of the kind that hid finding #1, and the service's remaining paths are read/update/delete only with no inserts. No further gaps found.

PR #184, opened 2026-07-26
test(mocks): make drizzle-mock.ts state-driven (single owner, unblocks 4 migration lanes) [KIM-443]

Oiranca:
## Problem

`tests/unit/mocks/drizzle-mock.ts` resolved every query by **call sequence** (`mockResolvedValueOnce` queued in the order calls were expected to happen). That model cannot represent a service that issues more than one DB call whose order is not statically fixed: concurrent `Promise.all` reads, a write followed by a read of the same table, or a conditional branch that skips a query all shift the queue and make an unrelated assertion fail. Worse, when the queue desynchronised the mock did not fail loudly — it handed the caller **another query's rows**, so a test could pass while asserting on data the service never fetched.

Four migration lanes were blocked on this at once, each re-deriving its own private fork of the helper (the usual anti-pattern: copy the mock from the most broken file, inherit its defect). Roughly **700k tokens** were burned across those lanes fighting the harness rather than the migration.

This PR makes the helper **state-driven and single-owner**: one in-memory store, queries resolved against it *at await time*.

## What changed

Two files, both under `tests/`. **No production code is touched.**

- `tests/unit/mocks/drizzle-mock.ts` (+1413 / -115) — rewritten around a `Map`-keyed in-memory row store plus an evaluator over Drizzle's own `SQL` condition AST (using `drizzle-orm` introspection symbols: `Column`, `Param`, `SQL`, `StringChunk`, `Table`, `getTableColumns`, `getTableName`, `is`).
- `tests/unit/mocks/drizzle-mock.test.ts` (new, 856 lines, 44 tests) — spec for the helper itself.

There is **no database here**: no pglite, no embedded or containerised Postgres, no SQL engine, no test DB, no live connection. `package.json` and `pnpm-lock.yaml` are unmodified; no dependency, script or postinstall hook was added.

## Public API

```ts
createStatefulDrizzleDb(): StatefulDrizzleDb   // hand to vi.mock('@/lib/db')
seed(data: Record<string, MockRow[]>): void    // replaces only the tables named
seedTable(table: TableRef, rows: MockRow[]): void
resetDb(): void                                // store + query log + injected failures
getRows(table: TableRef): MockRow[]            // snapshot, for post-write assertions
getQueryLog(): MockQueryLogEntry[]             // { op, table, rowCount }[] in order
failNextQuery(spec?: { op?: 'select'|'insert'|'update'|'delete'|'execute';
                       table?: TableRef; times?: number; error?: unknown }): void

type MockRow  = Record<string, unknown>
type TableRef = Table | string                 // Drizzle table object OR name
```

**Supported chain surface**

- `select()` / `select(fields)` / `selectDistinct()`, `.from`, `.innerJoin`/`.leftJoin` chained N times, `.where(cond | undefined)`, `.orderBy`, `.groupBy`, `.limit`, `.offset`, then `.execute()` or a plain `await`
- `insert().values()[.onConflictDoNothing() | .onConflictDoUpdate()][.returning()]`
- `update().set().where()[.returning()]` — awaitable without `.returning()`
- `delete().where()[.returning()]`
- `transaction(cb)` with rollback on throw

**Conditions:** `eq ne gt gte lt lte and or not inArray notInArray isNull isNotNull between notBetween like notLike ilike notIlike`, plus column-to-column comparisons.
**Aggregates:** `count()`, `count(col)`, `sum`, `avg`, `min`, `max`.

Schema defaults are applied on insert. `TableRef` normalises `'saved_games'`, `'savedGames'` and the table object to a single store entry; row keys accept either the Drizzle property name or the DB column name.

### For KIM-438 specifically — failure injection and rollback

The lane reported needing mid-transaction-failure simulation. Both capabilities exist:

- `failNextQuery({ op, table, times, error })` makes the *next* matching query (optionally the next `times` of them) reject with `error`, so a failure can be aimed at one specific statement inside a transaction rather than the whole callback.
- `transaction(cb)` snapshots the store copy-on-write and restores it if `cb` throws, so post-rollback state can be asserted with `getRows()`.

## Deliberately unsupported — these throw

Anything outside the surface above raises a named `[drizzle-mock] …` error rather than silently returning wrong rows. That is the entire point of the ticket: a mock that quietly returns the wrong data is worse than one that fails. Verified to throw: `rightJoin()`, `fullJoin()`, `.having()`, window functions (e.g. `row_number() over (…)`), non-column `groupBy`, unknown SQL operators, unparseable `orderBy`, and conditions referencing a table not part of the query.

**`db.execute(sql)` is the one documented exception and does *not* throw.** It is recorded in the query log and delegates to `executeMock`; when `executeMock` is unconfigured it resolves to `{ rows: [], rowCount: 0 }`. The store is not consulted, because arbitrary SQL cannot be evaluated without a SQL engine. Tests that rely on `db.execute` must configure `executeMock` explicitly.

Column-level constraints (`notNull`, unique indexes, FKs) are not enforced.

### Known gap for the migrating lanes

A **subquery passed as the right-hand operand of a comparison** — e.g. `inArray(col, db.select(...))` — is currently evaluated as an opaque value and yields **`[]` silently** instead of throwing. `exists(subquery)` does fail, but via a `Converting circular structure to JSON` error from the message renderer rather than a clean `[drizzle-mock]` message. No code in `lib/` or `app/` uses subqueries, CTEs or `EXISTS` today, so this is unreachable in the suites being migrated — but if a lane introduces one, it will get empty rows rather than a diagnostic. Tracked as follow-up; do not build on subquery support.

## Backward compatibility

The legacy sequence-driven mock is retained unchanged. All nine exports that `develop` actually consumes still exist: `selectMock`, `insertMock`, `updateMock`, `deleteMock`, `whereMock`, `SessionUser`, `createAdminSession`, `createMemberSession`, `createDrizzleQueryBuilder`. Symbols dropped from the parked forks have zero references on `develop`.

## Validation

Literal output on this branch:

```
pnpm test   →  Test Files  77 passed (77) /  Tests  1214 passed (1214)
typecheck   →  npx tsc --noEmit, exit 0, no output
lint        →  ✔ No ESLint warnings or errors
build       →  ✓ Compiled successfully, BUILD_EXIT=0
```

The pre-push hook ran full local CI and passed; `--no-verify` was not used.

**Test-count accounting:** develop 1170 → branch 1214, delta **+44**. `tests/unit/mocks/drizzle-mock.test.ts` is the only test file added or modified in the diff and contains exactly 44 `it(` blocks, so no pre-existing file's count changed.

The new assertions were mutation-tested rather than assumed meaningful: disabling conflict detection, `.offset()`, and the where-filter each broke the tests that claim to cover them; the helper was then restored byte-identical.

## Unblocks

Closes KIM-443

- KIM-438 https://linear.app/kimox-studio/issue/KIM-438
- KIM-439 https://linear.app/kimox-studio/issue/KIM-439
- KIM-440 https://linear.app/kimox-studio/issue/KIM-440
- KIM-441 https://linear.app/kimox-studio/issue/KIM-441

---

# Rebasing onto this branch — read before you start

## Take this file wholesale. Do not hand-merge.

If your branch carries its own commit to `tests/unit/mocks/drizzle-mock.ts`, **drop that commit before rebasing**, then resolve with `git checkout --ours tests/unit/mocks/drizzle-mock.ts`. Every functional capability the four forks added is already in this file.

**Git will not warn you if you skip this.** `git merge-tree` auto-merges the two versions with **no conflict markers at all** (exit 0). Nobody gets a conflict prompt. You get a clean-looking 2183-line file declaring 11 symbols twice, and it only surfaces at typecheck:

```
error TS2300: Duplicate identifier 'MockServiceError'.
error TS2323: Cannot redeclare exported variable 'createMockServiceError'.
error TS2393: Duplicate function implementation.
```

Verified on a clean checkout of this branch alone: 23 exports, zero repeats, no `export { ... }` lists, no `export default`, `pnpm typecheck` exit 0. The duplication is a rebase artefact, not a defect here — one lane rebased with its own frozen helper commit and hit the wall, another rebased without one and did not.

The colliding symbols are **not** the ones the forks added (`setUpdateFixture` / `setDeleteFixture` are unique names; `createDispatchingUpdateMock` / `createDispatchingDeleteMock` are file-local and never exported). They are the ones the old fixture file already shared with this one: `selectMock, insertMock, updateMock, deleteMock, whereMock, SessionUser, createAdminSession, createMemberSession, MockServiceError, createMockServiceError, createDrizzleQueryBuilder`.

## The fixture-queue API is dropped deliberately, not shimmed

`setFixture`, `setInsertFixture`, `setUpdateFixture`, `setDeleteFixture`, `setDefaultMockResponse`, `resetFixtures`, `getFixtureState`, `createDrizzleQueryBuilderWithDispatching`, `createTransactionAwareMockBuilder`, `createTransactionScopedBuilder` do **not** exist on `develop` — zero references anywhere, including develop's own copy of the mock. They only ever lived on the parked lanes, so there is nothing to keep backward-compatible, and a shim would preserve the very design that caused this ticket.

Your test files rewrite their fixture calls to seeded state. That is a real one-time cost per lane, stated up front rather than discovered mid-rebase.

The nine exports `develop` actually uses are preserved unchanged: `selectMock, insertMock, updateMock, deleteMock, whereMock, SessionUser, createAdminSession, createMemberSession, createDrizzleQueryBuilder`.

---

# Two sharp edges

Both are fine for sequential services and both are landmines otherwise:

1. **`db` and `adminDb` from two separate `createStatefulDrizzleDb()` calls share one module-level store.** A write through one is visible through the other.
2. **Rollback restores the whole store.** A write issued *outside* a transaction while that transaction was running is also discarded when it rolls back.

Do not write `Promise.all([db.transaction(...), db.insert(...)])`.

One practical note for anyone writing cross-run store comparisons: drop `defaultNow()` columns (`createdAt` / `updatedAt`) before comparing. They carry the wall clock and differ by milliseconds between runs. Assert their presence separately.

---

# Is the design actually proven?

Two independent answers, because "it has no consumers yet" is a fair thing to ask of 1400 lines of test infrastructure.

**A real service test file already runs on it.** KIM-441 rewrote `tests/unit/server/saved-games-service.test.ts` against the state-driven API: `8/8` on that file, `34/34` on `oir208-unified-events.test.ts`, full suite green, security-approved. That is the file that defeated three prior attempts under the sequence-based design — its `savedGamesJoinedQuery()` chains `.innerJoin().innerJoin()`, which the old mock could not express. The proof lives on that lane's branch rather than here, deliberately: migrating a lane's file into this PR would duplicate that lane's work.

**Order-independence under concurrency is proven here.** This is the criterion that motivated the whole redesign — `importMembersFromCsv` runs at `concurrencyLimit = 10`, so there is no call order to queue `mockResolvedValueOnce` against. The spec now runs the same 10 select-then-update flows (5 rows that exist, 5 that do not) under forward, reversed and shuffled arrival orders and asserts the final store and every per-flow result are identical across all three. Flows yield between their read and their write so writes genuinely land between other flows' reads, arrival orders are fixed permutations rather than `Math.random()` so any failure is reproducible, and the test asserts the three query-log signatures actually differ — otherwise "identical outcome" would be trivially true.

These tests were mutation-checked, not assumed. Injecting an order-dependence defect into the where-evaluator turns 5 tests red; resolving writes against a builder-construction-time snapshot instead of await time turns 1 red; applying a write column-by-column with a yield turns 1 red. The last two are caught **only** by the new tests — those two failure modes were previously undefended. One mutation (a cursor-dependent delay) did not flip the same-column-write test; probed in isolation, the delay genuinely does not change that schedule's observable outcome, so there was nothing to catch. Reported rather than papered over. The helper was restored byte-identical after every mutation (SHA256 verified).

---

# Validation

```
Test Files  77 passed (77)
     Tests  1231 passed (1231)
typecheck   exit 0, no diagnostics
lint        ✔ No ESLint warnings or errors
build       ✓ Compiled successfully, exit 0
```

Test counts: `develop` 1170 → this branch 1231. The entire delta is `tests/unit/mocks/drizzle-mock.test.ts` (61 tests), a new file covering new infrastructure. No pre-existing test file's count changed, and no test was deleted or weakened.

Commits: `be153f0` state-driven helper · `1e8b119` `execute()` failure injection fix · `ae1d254` its regression test · `52b9327` order-independence coverage.

### DIFF
--- a/tests/unit/mocks/drizzle-mock.ts
+++ b/tests/unit/mocks/drizzle-mock.ts
@@ -86,7 +86,8 @@
  * - `update(t).set(values).where(cond)[.returning(fields?)]`
  * - `delete(t).where(cond)[.returning(fields?)]`
  * - `transaction(cb)` — rolls the store back if `cb` throws
- * - `execute(sql)` — recorded in the query log, delegates to `executeMock`
+ * - `execute(sql)` — recorded in the query log, delegates to `executeMock`,
+ *   and honours `failNextQuery({ op: 'execute' })` like any other operation
  *
  * Conditions: `eq`, `ne`, `gt`, `gte`, `lt`, `lte`, `and`, `or`, `not`,
  * `inArray`, `notInArray`, `isNull`, `isNotNull`, `between`, `notBetween`,
@@ -433,6 +434,14 @@ const queryLog: MockQueryLogEntry[] = []
 type FailureSpec = { op?: QueryOp; table?: string; times: number; error: unknown }
 const failures: FailureSpec[] = []
 
+/**
+ * Table key used for raw `db.execute(sql\`...\`)` calls, in the query log and
+ * when matching injected failures. Raw SQL names no table the mock can parse,
+ * so `execute` gets the empty key: op-only and unfiltered failures match it,
+ * table-scoped ones deliberately do not.
+ */
+const EXECUTE_TABLE_KEY = ''
+
 /** Normalise any table reference to a single store key. */
 function tableKey(ref: TableRef): string {
   const name = typeof ref === 'string' ? ref : getTableName(ref)
@@ -496,6 +505,17 @@ export function getQueryLog(): MockQueryLogEntry[] {
  * failNextQuery({ op: 'update', table: profiles })
  * // the service's catch-block path is now exercised
  * ```
+ *
+ * Applies to every operation in {@link QueryOp}, `execute` included:
+ *
+ * ```typescript
+ * failNextQuery({ op: 'execute' })
+ * // the next db.execute(sql`...`) rejects; the one after it resolves normally
+ * ```
+ *
+ * `table` cannot be used together with `op: 'execute'` — a raw-SQL statement
+ * names no table the mock can parse, so such a spec could never match. That
+ * combination throws here rather than sitting in the queue and never firing.
  */
 export function failNextQuery(spec: {
   op?: QueryOp
@@ -503,6 +523,12 @@ export function failNextQuery(spec: {
   times?: number
   error?: unknown
 } = {}): void {
+  if (spec.op === 'execute' && spec.table !== undefined) {
+    throw mockError(
+      "failNextQuery({ op: 'execute' }) cannot be scoped by table — raw SQL names no table the mock can " +
+        'parse. Drop the `table` field to fail the next execute regardless of the statement.',
+    )
+  }
   failures.push({
     op: spec.op,
     table: spec.table === undefined ? undefined : tableKey(spec.table),
@@ -1535,10 +1561,28 @@ export function createStatefulDrizzleDb() {
       set: vi.fn((values: MockRow) => new MockUpdateBuilder(table, values)),
     })),
     delete: vi.fn((table: Table) => new MockDeleteBuilder(table)),
-    execute: vi.fn((...args: unknown[]) => {
-      queryLog.push({ op: 'execute', table: '', rowCount: 0 })
+    /**
+     * Raw SQL passthrough. The store is never consulted (arbitrary SQL cannot be
+     * evaluated without a SQL engine), so the result comes from `executeMock`,
+     * defaulting to `{ rows: [], rowCount: 0 }`.
+     *
+     * Failure injection applies here exactly as it does to the other operations:
+     * `failNextQuery({ op: 'execute' })` — or an unfiltered `failNextQuery()` —
+     * makes the next call reject *before* `executeMock` is reached, and the
+     * rejected call is left out of the query log.
+     *
+     * A raw-SQL statement names no table the mock can parse, so `execute` is
+     * matched against {@link EXECUTE_TABLE_KEY} (the empty key) rather than a
+     * real one: a table-scoped failure such as `failNextQuery({ table: rooms })`
+     * deliberately does not fire on `execute`. The one combination that could
+     * never fire — an explicit `{ op: 'execute', table }` — is rejected up front
+     * by {@link failNextQuery} rather than silently ignored.
+     */
+    execute: vi.fn(async (...args: unknown[]) => {
+      consumeFailure('execute', EXECUTE_TABLE_KEY)
+      queryLog.push({ op: 'execute', table: EXECUTE_TABLE_KEY, rowCount: 0 })
       const result = executeMock(...args)
-      return Promise.resolve(result ?? { rows: [], rowCount: 0 })
+      return result ?? { rows: [], rowCount: 0 }
     }),
     /**
      * Runs `callback` against the same store. If it throws, every write made
--- a/tests/unit/mocks/drizzle-mock.ts
+++ b/tests/unit/mocks/drizzle-mock.ts
@@ -1,79 +1,143 @@
 /**
- * Shared Drizzle query builder mock factory and state for service tests
+ * Shared Drizzle mock infrastructure for service-layer unit tests.
  *
- * This helper provides reusable mock infrastructure for testing Drizzle-based
- * service layers across multiple test files. It handles chainable query builders,
- * transaction wrappers, and common CRUD operations (select, insert, update, delete).
+ * This file exposes TWO independent mock flavours. Pick one per test file —
+ * they do not interact.
  *
- * ## Why This Exists
+ * ─────────────────────────────────────────────────────────────────────────────
+ * 1. STATE-DRIVEN MOCK (`createStatefulDrizzleDb`) — preferred (KIM-443)
+ * ─────────────────────────────────────────────────────────────────────────────
  *
- * Unit tests for Drizzle services need to mock the query builder chain:
- * `.select().from().where()`, `.insert().values().returning()`, etc.
- * Without a shared helper, each test file had to hand-wire all these chains locally,
- * creating duplicate mock logic and making it tedious to set up new test files.
+ * An in-memory JS store (a `Map` keyed by table name) that answers every query
+ * by *inspecting the builder chain* (table, joins, where-condition, order,
+ * limit) and evaluating it against the seeded rows. There is NO reliance on
+ * call ordering, so it is correct for services that issue concurrent or
+ * interleaved queries (e.g. `importMembersFromCsv`, which runs with
+ * `concurrencyLimit = 10`).
  *
- * ## How to Use in a Test File
+ * There is no real database here: no pglite, no embedded/containerised
+ * Postgres, no SQL engine, no connection. Conditions built with `eq`, `and`,
+ * `or`, `inArray`, `gte`, … are decoded from Drizzle's own `SQL` AST and
+ * evaluated against plain JS objects.
  *
- * 1. Import the mock state and factory:
- *    ```typescript
- *    import {
- *      createDrizzleQueryBuilder,
- *      selectMock,
- *      insertMock,
- *      updateMock,
- *      deleteMock,
- *      createAdminSession,
- *      createMemberSession,
- *    } from '@/tests/unit/mocks/drizzle-mock'
- *    ```
+ * ### Usage
  *
- * 2. Set up the vi.mock call (must be at the top level of the test file):
- *    ```typescript
- *    vi.mock('@/lib/db', () => ({
- *      getDrizzleDb: vi.fn(() => createDrizzleQueryBuilder()),
- *      getDrizzleAdminDb: vi.fn(() => createDrizzleQueryBuilder()),
- *    }))
- *    ```
+ * ```typescript
+ * import {
+ *   createStatefulDrizzleDb,
+ *   seed,
+ *   resetDb,
+ *   getRows,
+ * } from '@/tests/unit/mocks/drizzle-mock'
  *
- * 3. In beforeEach, reset the mocks:
- *    ```typescript
- *    beforeEach(() => {
- *      vi.resetModules()
- *      vi.clearAllMocks()
- *    })
- *    ```
+ * // vi.mock() is hoisted, so it must live in the test file itself (see the
+ * // note at the bottom of this header) — but the factory can be shared.
+ * vi.mock('@/lib/db', () => ({
+ *   getDrizzleDb: vi.fn(() => createStatefulDrizzleDb()),
+ *   getDrizzleAdminDb: vi.fn(() => createStatefulDrizzleDb()),
+ * }))
  *
- * 4. In individual tests, configure the mock responses:
- *    ```typescript
- *    selectMock.mockResolvedValue([{ id: 'row-1', name: 'Example' }])
- *    // or
- *    insertMock.mockResolvedValue([{ id: 'new-row' }])
- *    // or
- *    updateMock.mockResolvedValue([{ id: 'updated-row' }])
- *    // or
- *    deleteMock.mockResolvedValue([{ id: 'deleted-row' }])
- *    ```
+ * beforeEach(() => {
+ *   resetDb()
+ *   vi.clearAllMocks()
+ * })
+ *
+ * it('returns only the member own reservations', async () => {
+ *   seed({
+ *     reservations: [
+ *       { id: 'r-1', userId: 'member-1', tableId: 't-1', date: '2026-08-01' },
+ *       { id: 'r-2', userId: 'member-2', tableId: 't-1', date: '2026-08-01' },
+ *     ],
+ *     tables: [{ id: 't-1', roomId: 'room-1', name: 'Table 1' }],
+ *   })
+ *
+ *   const result = await listReservationsForSession(createMemberSession())
+ *
+ *   expect(result).toHaveLength(1)
+ *   // writes are visible to later reads in the same test:
+ *   expect(getRows('reservations')).toHaveLength(1)
+ * })
+ * ```
+ *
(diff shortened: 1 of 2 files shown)


tests/unit/mocks/drizzle-mock.ts
```diff
@@ -208,52 +360,1198 @@ function createSelectWhereMock() {
...
+export function createStatefulDrizzleDb() {
+  return {
+    select: vi.fn((fields?: FieldMap) => new MockSelectBuilder(fields, false)),
+    selectDistinct: vi.fn((fields?: FieldMap) => new MockSelectBuilder(fields, true)),
+    insert: vi.fn((table: Table) => new MockInsertBuilder(table)),
+    update: vi.fn((table: Table) => ({
+      set: vi.fn((values: MockRow) => new MockUpdateBuilder(table, values)),
+    })),
+    delete: vi.fn((table: Table) => new MockDeleteBuilder(table)),
+    execute: vi.fn((...args: unknown[]) => {
```
Oiranca: `failNextQuery()` exposes `op: 'execute'`, and `QueryOp` includes `'execute'`, but this `execute()` path never calls `consumeFailure('execute', ...)`. That means a test can write `failNextQuery({ op: 'execute' })` expecting to exercise a raw-SQL failure path, but the mock will still resolve through `executeMock` or the default `{ rows: [], rowCount: 0 }` response.

Please either wire failure injection into this path, e.g. consume an execute failure before logging/delegating, or remove `'execute'` from the supported `QueryOp` surface. Add a regression test with `failNextQuery({ op: 'execute' })` that proves the next `db.execute(...)` rejects and the following one succeeds.

Oiranca (commented):
Reviewed for best practices, security, and test coverage. I found one test-infrastructure contract issue that should be fixed so error-path tests cannot get false confidence.

tests/unit/mocks/drizzle-mock.ts
Oiranca: ✅ Fixed in `1e8b119`, regression test in `ae1d254`.

You were right, and this was the exact failure class this PR exists to eliminate — a test believing it exercises an error path while the mock quietly succeeds. It was not hypothetical either: `reservations-service.ts` replaced the `mark_no_show_reservations` RPC with a raw ``db.execute(sql`...`)``, so the reservations lane's failure-path tests would have landed squarely in it.

We took the first option you offered (wire injection in) rather than shrinking `QueryOp`, because that capability is genuinely needed by the reservations migration.

```ts
// before
execute: vi.fn((...args: unknown[]) => {
  queryLog.push({ op: 'execute', table: '', rowCount: 0 })
  const result = executeMock(...args)
  return Promise.resolve(result ?? { rows: [], rowCount: 0 })
}),

// after
execute: vi.fn(async (...args: unknown[]) => {
  consumeFailure('execute', EXECUTE_TABLE_KEY)
  queryLog.push({ op: 'execute', table: EXECUTE_TABLE_KEY, rowCount: 0 })
  const result = executeMock(...args)
  return result ?? { rows: [], rowCount: 0 }
}),
```

Three details worth recording:

- **`async` is load-bearing.** The other four ops call `consumeFailure` inside a private `async run()`, so an injected error surfaces as a rejected promise. Making `execute` async gives it identical semantics rather than a synchronous throw. `executeMock` is still invoked synchronously within the call, so call-ordering assertions are unaffected.
- **Log ordering had to be swapped.** `consumeFailure` now precedes `queryLog.push`, so a failed execute is not logged — matching select/insert/update/delete exactly. A successful one still logs `{ op: 'execute', table: '', rowCount: 0 }`.
- **`table` cannot be matched for raw SQL**, so execute uses a new `EXECUTE_TABLE_KEY = ''`. A table-scoped-but-op-agnostic `failNextQuery({ table: profiles })` deliberately does not fire on execute. The one combination expressing explicit intent that could never fire — `failNextQuery({ op: 'execute', table })` — now throws at `failNextQuery()` call time rather than sitting queued and silently never firing. Silently ignoring the field would have recreated the same false-confidence class. Documented in code comments on `execute`, on `failNextQuery`, and on the constant.

The legacy `createDrizzleQueryBuilder().execute` was left alone — the failure mechanism belongs to the state-driven mock only.

**Regression coverage:** 12 tests in a new `describe('error injection on raw db.execute()')` block, including the both-halves case you asked for (next `db.execute` rejects, the one after succeeds). They were mutation-checked rather than assumed: reverting the helper to its pre-fix state turns **10 of the 12 red**, while all 44 pre-existing tests stay green. The 2 survivors are intentional — they pin the deliberate design decisions above (table-scoped spec skips execute; unconfigured `executeMock` returns the documented default), which the pre-fix code also satisfied.

Gate after the fix: `Test Files 77 passed (77) / Tests 1226 passed (1226)`, typecheck exit 0, `✔ No ESLint warnings or errors`, build exit 0.

Oiranca:
## Notes for the four rebasing lanes (KIM-438 / 439 / 440 / 441)

### The "duplicate exports" wall is a rebase artefact, not a defect in this branch

KIM-438 hit duplicate-export errors rebasing onto this branch; KIM-439 rebased onto the same branch with no such problem. The difference is that KIM-438 carried its own frozen local helper commit (`4c7434d`) into the rebase and KIM-439 had no equivalent.

Verified on a clean checkout of this branch alone: **23 exports, zero repeats** (no `export { ... }` lists, no `export default`), `pnpm typecheck` exit 0.

The collision is created by combining the two versions, and it is worth knowing exactly how, because **git does not warn you**. `git merge-tree` auto-merges the two files with **no conflict markers at all** (exit 0). Nobody gets a conflict prompt. You get a clean-looking 2183-line file declaring 11 symbols twice, and it only surfaces at typecheck:

```
error TS2300: Duplicate identifier 'MockServiceError'.
error TS2323: Cannot redeclare exported variable 'createMockServiceError'.
error TS2393: Duplicate function implementation.
```

The colliding symbols are **not** the ones `4c7434d` added — `setUpdateFixture` / `setDeleteFixture` are unique names, and `createDispatchingUpdateMock` / `createDispatchingDeleteMock` are file-local and never exported. They are the ones that lane's *base* fixture file shares with this one: `selectMock, insertMock, updateMock, deleteMock, whereMock, SessionUser, createAdminSession, createMemberSession, MockServiceError, createMockServiceError, createDrizzleQueryBuilder`.

**Resolution: drop `4c7434d` before rebasing, and take this branch's version of the file wholesale** (`git checkout --ours tests/unit/mocks/drizzle-mock.ts`). Do not hand-merge — every functional capability those forks added is already in this file.

### Transaction scoping and rollback are confirmed by probe

The root cause behind the old helper's atomicity failures was that `TransactionScope` copied only `tableFixtures`, so `tx.update(...).set(...).where(...).returning()` inside a transaction read global state, returned `[]`, and the service threw a misleading 500. That does not reproduce here. Verified:

**Per-table writes inside one transaction** — the club-events shape, 4 tables, one `db.transaction()`:

```
updateEventsReturning:       [{"id":"...001","title":"New title"}]   <- NOT [], destructuring gives a row
insertEventsReturning:       [{"id":"...007","title":"Inserted"}]
deleteBlocks/Equipment/Events without .returning(): []
updateReservationsReturning: [{"id":"...006","status":"cancelled"}]
events(matched) ["Hit"] vs reservations(unmatched) []   <- distinct per-table results in the same tx
```

**Rolled-back transaction leaves no trace on a later read:**

```
select(events).title:        ["Original title"]     getRows(events).title: ["Original title"]
select(eventRoomBlocks):     length 0               getRows(event_room_blocks): []
select(reservations).status: ["pending"]
plain business-logic throw -> eventRoomBlocks: [], eventEquipment: []   <- no orphaned rows
a DELETE inside the tx is restored; a builder constructed before the tx and awaited after also sees rolled-back state
mirror case commits: events ["Committed title"], blocks 1, reservations ["cancelled"]
```

Combined with `failNextQuery({ op, table, times, error })` firing on the intended write rather than an earlier one, this is enough to un-`skip` the five atomicity tests waiting on it.

### An UPDATE that matches no row cannot create one

Verified with 12 probes, because an external review found a test mock elsewhere that fabricated a row on UPDATE — modelling what the calling code expected instead of what Postgres does — and that hid a genuine production bug.

With and without `.returning()`, on a seeded table with no match, on an empty table, with no `.where()` at all, inside a transaction, and with `.set()` carrying the primary key of a nonexistent row: always `[]`, store untouched. It is structurally impossible — `MockUpdateBuilder.run()` only filters existing rows, and the single `stored.push()` in the file is in `MockInsertBuilder`. Counter-proof that this is not a "never writes" false negative: an UPDATE with no `.where()` on a *seeded* table correctly updates all rows. `onConflictDoUpdate` with a target that matches nothing performs a clean INSERT of the `values` row without applying the `set` payload.

### Two behaviours to know before writing tests

- `db` and `adminDb` from two separate `createStatefulDrizzleDb()` calls **share one module-level store**.
- Rollback restores the **whole** store, so a write issued outside the transaction while it was running is also discarded. This does not affect services that are sequential inside one transaction, but do not write `Promise.all([db.transaction(...), db.insert(...)])`.

### Honest scope note

This helper is not yet load-bearing: `createStatefulDrizzleDb` / `seed` / `resetDb` / `failNextQuery` have exactly one consumer, their own spec. The six real test files still use the legacy compatibility shim, which is preserved unchanged. Migrating those files onto the state-driven API is the work that actually validates the design, and it belongs to the individual lanes.

PR #185, opened 2026-07-26
feat(auth): cut login/session path over to Auth.js (KIM-433)

Oiranca:
## Summary

Cuts the login/session path over to Auth.js (Credentials provider + bcrypt, JWT strategy), which existed since KIM-416/PR #170 but was never the real session path — the app used Supabase Auth end-to-end until this change.

- `app/api/auth/{login,logout,activate,recover}/route.ts` — no longer use `@/lib/supabase/server`; session issuance now goes through Auth.js.
- `app/api/auth/callback/route.ts` — **deleted**. It was Supabase-PKCE-only (`exchangeCodeForSession`); there is no Credentials-provider equivalent, and no other code references this route.
- `lib/server/auth/auth.ts` — session read now via Auth.js `auth()` + Drizzle/Neon `profiles` lookup (role/is_active re-verified fresh from the DB on every call, not trusted from the JWT).
- `lib/server/auth/auth-service.ts` — `login()`/`logout()`/`activateAccount()`/`recoverAccount()` now issue Auth.js sessions. `register()` is **untouched, out of scope** — it still creates real Supabase Auth sessions and is tracked separately (KIM-444).
- `lib/authjs/credentials-user.ts` — now looks up `profiles` via Drizzle, keyed on the uniquely-indexed `auth_email` column (was previously scaffolded against the non-unique `email` column — fixed as part of this PR, see below).
- `lib/authjs/config-edge.ts` (new) — Edge-safe shared session/cookie config (no Credentials provider, no DB access), explicit 7-day `session.maxAge` (previously would have silently defaulted to NextAuth's 30 days).
- `middleware.ts` — reads no session at all now (a dead `getToken()` call from an earlier commit in this branch was removed — its result was never used for gating on this branch *or* on `develop`). Page-level and API-level route protection is unaffected by this: both continue to run through `getSessionFromServerCookies()`/`requireAuth`/`requireAdmin` from `lib/server/auth/auth.ts`, now backed by `auth()` + a fresh Drizzle role/is_active check instead of Supabase — e.g. `app/[locale]/admin/page.tsx` still redirects unauthenticated/non-admin users correctly, verified unchanged on this branch. What did **not** change, then or now, is that this protection has only ever run at the page/route level, never in `middleware.ts` itself.
- `lib/server/shared/security-edge.ts` — **untouched, zero diff vs develop.** Its only `@supabase/ssr` reference is a type-only import (`CookieOptionsWithName`); `getSupabaseCookieOptions()` still has a live runtime consumer in `lib/supabase/server.ts` because `register()` still needs it. This is the last Supabase reference on the auth surface — noted here explicitly rather than implying the cutover removed all Supabase dependencies.
- `lib/supabase/` and `@supabase/*` packages are deliberately **not removed** — that's KIM-422, gated on this PR merging first.

## A real bug found and fixed during review (not cosmetic)

An early version of this branch restored a login identity-drift guard (`develop` had `if (data.user.id !== credentialProfile.id) serviceError(401)`) using a pattern that called `signIn()` then immediately read back the session via `auth()` to get the authenticated id. That pattern is **structurally broken**: `next-auth`'s `signIn()` builds its internal request from `next/headers`' `headers()` — a frozen snapshot of the *incoming* request — while writing the new session cookie to a separate mutable `cookies()` jar; zero-arg `auth()` also resolves via that same frozen snapshot. In the ordinary case (no pre-existing session cookie), the readback always returned `null`, meaning **valid credentials would have gotten a 401 while a genuine, valid session cookie still rode on that 401 response.**

Fixed (commit `244bcbc`): `login()`/`activateAccount()`/`recoverAccount()` now resolve and verify the authenticated identity via `verifyCredentials()` **before** ever calling `signIn()`, and compare it against the independently-resolved profile. On mismatch, `signIn()` is never called, so no session cookie is ever issued — no cleanup needed, no window where a cookie rides on a failure response. The reasoning is documented in `lib/server/auth/auth-service.ts`'s `establishAuthJsSession()` doc comment, citing the specific `next-auth` source files, so a future change can't reintroduce the pattern without seeing why it's wrong. Regression trip-wire tests added in `tests/unit/server/auth-service.test.ts` assert `signIn`/`auth()` are never called on the paths where they must not be.

## Test count — stated honestly, not as parity

`develop`: **1170** tests. This branch: **1178** tests. Net **+8**, composed of:
- **+11** new file `tests/unit/authjs/config-edge.test.ts` (previously-untested Auth.js edge cookie/session policy).
- **+1** `tests/unit/server/auth-service.test.ts` (29 → 30: identity-drift guard rewrite + regression trip-wire).
- **−3** `tests/unit/app/api/auth-routes.test.ts` (18 → 15): tests for the deleted `callback` route's PKCE exchange/redirect-sanitization behavior. Behavior was removed, not relocated — verified no other route references `/api/auth/callback`.
- **−1** `tests/unit/app/middleware.test.ts` (3 → 2): a test asserting middleware passed Supabase cookie options into `createServerClient` — that mechanism is gone. The underlying cookie-security-flag behavior it exercised is still covered (4×) in `tests/unit/server/security.test.ts`, and the *replacement* Auth.js cookie policy is covered by the new `config-edge.test.ts`.

Both counts (1170 and 1178) were computed directly via `grep -cE "^\s*(it|test)\(" ` across every file in `tests/` on each branch, not taken from any tool's self-report.

## Literal validation output

```
Test Files  77 passed (77)
Tests  1178 passed (1178)
Start at  19:12:16
Duration  25.88s (transform 841ms, setup 2.17s, collect 2.56s, tests 9.01s, environment 3.42s, prepare 2.30s)
```

`pnpm typecheck`: exit 0, `✓ Route types generated successfully`, no `tsc` errors.
`pnpm lint`: `✔ No ESLint warnings or errors`.
`pnpm build`: exit 0, `✓ Compiled successfully`, all 42 static pages generated.
Local pre-push hook (full CI re-run) passed on push: `━━━ CI local pasado ✓ ━━━`.

All tests are mock-only — no real DB connection, no containerized/embedded Postgres, no live HTTP session, per repo-wide test rules. This is also the reason a live login round-trip is not covered by anything above (see next section).

## What has NOT been verified — manual step required before merge

**A live login round-trip has not been performed.** This is not an environment limitation to be resolved later — it's a permanent property of this repo's test rules (tests never touch a real DB), so it will never show up as a passing automated test.

KIM-432 (already merged, PR #177) seeded a local Neon dev DB with real bcrypt password hashes: `admin@devseed.local` (admin role) and `member1/2/3@devseed.local` (member role), password set via the `DEV_SEED_PASSWORD` constant in `scripts/seed.ts` (name only — not reproduced here).

**Before merging this PR, please perform a manual login against a seeded local Neon using `admin@devseed.local`** to confirm the full request path (route handler → `verifyCredentials()` → Drizzle → bcrypt → Auth.js session issuance) works end to end, not just at the unit level.

> ⚠️ **That seeded-admin login alone is NOT sufficient, and on its own would have produced a false green.** Code review (Codex) found that `activateAccount()`/`recoverAccount()` never wrote the new password to `profiles.password_hash` — the credential source `verifyCredentials()` reads — so activation appeared to succeed while every subsequent login failed. The seeded admin would have passed this test regardless, because `scripts/seed.ts` writes `password_hash` directly; only users who went through activation were broken. **The manual check must therefore also cover: activate a fresh user with a new password, then log in as that user** (and the same for the recovery flow). Fixes for this and for a related issue (sessions being issued for inactive accounts) are being pushed to this branch — see the review threads.

**Caveat on how much that manual test will actually cover:** `auth-service.ts`'s broader business logic — `register()`, and roughly 10 other `getAdminDb()`/`getDb()` call sites for profile reads — is still on the Supabase seam and explicitly out of scope for this PR (tracked as a new, separate issue, KIM-444). A successful Neon credential check on this branch will still be followed by Supabase-backed profile-read code paths elsewhere in the request lifecycle for many flows. This PR is the login/session cutover specifically, not full end-to-end migration of everything auth-adjacent.

## Other notes for reviewers

- `middleware.ts` no longer refreshes any session cookie (the previous Supabase `getUser()` call, which existed purely to refresh the cookie before expiry, was removed with no Auth.js equivalent — none is needed for the JWT strategy). This is safe **today** only because `app/api/auth/register/route.ts` unconditionally returns 410 (self-registration disabled), making `register()`'s real Supabase session issuance unreachable dead code. If KIM-444 or any future issue re-enables self-registration, this transitional gap (no refresh mechanism for a Supabase-issued session) needs revisiting.
- `session.maxAge` is now an explicit 7 days (was an implicit NextAuth default of 30 days) — a deliberate choice made and documented as part of this PR, not carried over from develop (which had its own, differently-configured Supabase session lifetime).

## Review status

Reviewed twice by security-reviewer. First pass found a CRITICAL issue (the signIn-then-auth() readback bug described above) — fixed, then independently re-verified against the actual installed `next-auth`/`@auth/core` source (not against a summary) in a second pass. Final verdict: **APPROVE**, CRITICAL finding confirmed fixed.

## Merge

Not merging this PR myself. Per project process, the user merges manually after reviewing — and per current instruction, only once review comments are resolved and the user explicitly says to merge, independent of test suite status.


lib/server/auth/auth-service.ts
```diff
@@ -437,9 +497,36 @@ export async function activateAccount(input: { token: unknown; password: unknown
...
+  // route handler doesn't need any auth-client construction of its own.
+  // Account activation has already succeeded at this point regardless of
+  // whether this sign-in attempt succeeds — `signInFailed` lets the route
+  // handler report that distinction to the client (same behavior as before).
+  //
+  // Identity is resolved and compared BEFORE any session is established —
+  // see `establishAuthJsSession()`'s doc comment (KIM-433) for why a
+  // same-request signIn()-then-auth() readback cannot be relied on instead.
+  const authEmail = profile.auth_email ?? profile.email ?? ''
+  const verifiedUser = await verifyCredentials(authEmail, parsed.data.password)
```
Oiranca, line 598: This verification now reads the password from Neon's `profiles.password_hash`, but the activation flow above only updates the password through Supabase Auth (`updateAuthUserById`) and then updates profile metadata (`is_active`, `active_from`, `psw_changed`). It never writes the new bcrypt hash into `profiles.password_hash`. That means activation can successfully change the Supabase password and mark the account active, then `verifyCredentials(authEmail, password)` will fail against stale/null Neon state, and future Auth.js logins with that new password will fail as well. The same problem exists in `recoverAccount()` after its `updateAuthUserById` call. Please make activate/recover update the Auth.js credential source as part of the password change, and add regression coverage proving the new password can authenticate after each flow.

lib/authjs/credentials-user.ts
```diff
@@ -40,43 +46,47 @@ interface ProfileRow {
...
     }
 
-    const passwordMatches = await bcrypt.compare(password, row.password_hash)
+    const passwordMatches = await bcrypt.compare(password, row.passwordHash)
 
     if (!passwordMatches) {
       return null
     }
 
     return {
```
Oiranca, line 88: `verifyCredentials()` returns a user even when `row.isActive` is false. Most app routes re-check `isActive` later in `getSessionUser()`, but the Credentials provider is still allowed to issue a valid Auth.js session token for a suspended profile, especially once `/api/authjs/*` is enabled or if the legacy Supabase profile row and Neon profile row drift during this transition. Authentication should fail before session issuance for inactive profiles. Please return `null` when `!row.isActive`, and add tests for direct Credentials authorization and `/api/auth/login` asserting `signIn()` is not called for inactive Neon rows.

Oiranca (commented):
Auth cutover review: found two blocking auth/session issues around password-hash consistency and inactive-account session issuance.

lib/server/auth/auth-service.ts
Oiranca: ✅ Fixed in 6db5ac5 / 8f0b59e. `activateAccount()` and `recoverAccount()` now write the new bcrypt hash through the Drizzle/Neon seam (`profiles.password_hash`) in addition to the Supabase Auth `updateAuthUserById` call, and a zero-row Drizzle `UPDATE` is now treated as a failed write (surfaces as an error and rolls back the activation/recovery token) rather than silently succeeding.

Regression coverage added in b35f815:
- `tests/unit/server/auth-activation.test.ts`: "activation succeeds and writes password hash to Drizzle seam" and "treats zero-row Drizzle password hash update as a failed write and rolls back token"
- `tests/unit/server/auth-recovery.test.ts`: equivalent pair for `recoverAccount()`

Both were proven to fail against the pre-fix code (reverted to commit before 6db5ac5/8f0b59e) before being confirmed green with the fix in place.

lib/authjs/credentials-user.ts
Oiranca: ✅ Fixed in 6db5ac5. `verifyCredentials()` now returns `null` when `!row.isActive`, before `bcrypt.compare()` is ever invoked, so a suspended profile cannot be issued an Auth.js session token regardless of correct credentials.

Regression coverage added in b35f815 in `tests/unit/authjs/credentials-user.test.ts`:
- "returns null when profile is inactive (isActive: false)" — asserts `bcrypt.compare` is not called at all for an inactive row
- "returns null uniformly for inactive profile, matching 'no user found' and 'wrong password' signals" — defense-in-depth check that the inactive path doesn't leak account-existence info via a different return shape

Both were proven to fail against the pre-fix code (bcrypt.compare was being invoked and a user object returned) before being confirmed green with the fix in place.

